PR #78, opened 2022-11-30
feature: Add plotting functions (histogram, scatterplot, boxplot) for exploratory analysis.

sahautal:


nialov:
`pandas` already provides quite well-integrated plotting capabilities for these types of plots. E.g. `pandas.DataFrame.plot.bar`, `pandas.DataFrame.plot.hist`, `pandas.DataFrame.plot.scatter`. The wanted column from the dataframe can be passed as an argument to the functions. Is there a reason not to use these? At a high level choosing the columns by name rather than index is also clearer although then there has to be checks that the input dataframe columns have known names.

Of course if customization of the plots at a "deep" level is required at some point it is probably best to do everything as low level as possible and you want to separate the `pandas.DataFrame` logic from plotting then working with just `numpy` arrays works well.

nmaarnio:
I agree with @nialov , pandas plots would make more sense here. I do not think there was really any reason for the input to be index instead of column name (I was not in the project in the early planning phase). As a general comment for all of us EIS Toolkit developers, the planned interfaces (and implementations/libraries used) are not strict and can be challenged by the developer if they see a better solution.

As another general remark about plotting functions, we need to plan in the near future how we wish to implement them. There are questions about plotting libraries, output types, styles and parameters that should be discussed to make a consistent plan. I think the end users want either high degree of control of their graphs or just the right kind of graph for each different result type.

nmaarnio:
As the plans for developing plotting functions have changed somewhat and a new branch and issue were created for this purpose, I will close this pull request. The new relevant issue is [this one](https://github.com/GispoCoding/eis_toolkit/issues/117).

PR #79, opened 2022-11-30
Gtk/fuzzy overlay

jtorppa:
Generated functions for fuzzy overlay computation.


nmaarnio (dismissed):
Here are some general comments:
- Remember to run "invoke lint" before committing. The linter will point you where in your code there are formatting errors, and in some cases autoformat
- Please create also test functions for cases where the function works as expected
- The test notebook should be named for example "testing_fuzzy.ipynb". Also please follow the look and structure of testing_clip.ipynb if you wish to have a notebook for testing. However, I think for these functions normal unit tests could suffice
- Remember to write return type annotations, even if Python is possible to write without them
- The specific comments I made within the code applied often for multiple places, but I did not mark each of the occurrences
- I am not sure what should be used as nodata in these functions, but I think it is not too bad if we leave it as a TODO

nmaarnio:
This PR has been hanging here for quite some time. Could we get this finished and merged sometime soon @jtorppa or perhaps @nialov? I think it'd be pretty quick to get it in shape.

nialov:
I can clean it up and add some basic sanity check tests. However, I probably cannot guarantee that the methods are correct for the implementation. Up to @jtorppa.

eis_toolkit/prediction/fuzzy_overlay.py
```diff
@@ -0,0 +1,112 @@
...
+            Data points should be in the range [0, 1].
+
+    Returns:
+        2D Numpy array with the result of the 'and' overlay operation. Values are in range [0, 1].
+
+    Raises:
+        InvalidParameterValueException: If data values are not in range [0, 1].
+    """
+    if any(band_data.min() < 0 or band_data.max() > 1 for band_data in data):
+        raise InvalidParameterValueException("All data must be in range [0, 1]")
```
nialov, line 22: This check should be refactored to a simple check function so that it can be reused.

```python
def _check_input_data(data):
    if any(band_data.min() < 0 or band_data.max() > 1 for band_data in data):
        raise InvalidParameterValueException("All data must be in range [0, 1]")
```

```suggestion
    _check_input_data(data=data)

```

eis_toolkit/prediction/fuzzy_overlay.py
```diff
@@ -0,0 +1,112 @@
...
+        InvalidParameterValueException: If data values or gamma are not in range [0, 1].
+    """
+    if any(band_data.min() < 0 or band_data.max() > 1 for band_data in data):
+        raise InvalidParameterValueException("All data must be in range [0, 1]")
+
+    if gamma < 0 or gamma > 1:
+        raise InvalidParameterValueException("The gamma parameter must be in range [0, 1]")
+
+    fuzzy_sum = data.sum(axis=0) - np.prod(data, axis=0)
+    fuzzy_product = np.prod(data, axis=0)
```
nialov, line 111: Why not reuse already defined functionality for sum and product?

```suggestion
    fuzzy_product = product_overlay(data=data)
```

tests/prediction/fuzzy_test.py
```diff
@@ -0,0 +1,64 @@
...
+    assert np.allclose(result, [[1.0, 0.65, 0.28], [0.82, 0.96, 0.85], [1.0, 0.36, 1.0]])
+
+
+def test_gamma_overlay():
+    """Test that GAMMA overlay works as expected."""
+    result = gamma_overlay(data=RASTERS_DATA, gamma=0.6)
+    assert np.array_equal(
+        np.around(result, decimals=4), [[0.8670, 0.3616, 0.0974], [0.5335, 0.7626, 0.5960], [0.9587, 0.1495, 0.7579]]
+    )
+
```
nialov, line 45: Simple sanity check for the mathematics based on the definitions in literature. Maybe useful (probably not).

```suggestion
def test_gamma_overlay_equivalent_product():
    """Test that GAMMA overlay is equal to product with 0.0 gamma."""
    result_gamma = gamma_overlay(data=RASTERS_DATA, gamma=0.0)
    result_product = product_overlay(data=RASTERS_DATA)
    np.testing.assert_array_equal(result_gamma, result_product)

```

tests/prediction/fuzzy_test.py
```diff
@@ -0,0 +1,64 @@
...
+def test_sum_overlay():
+    """Test that SUM overlay works as expected."""
+    result = sum_overlay(data=RASTERS_DATA)
+    assert np.allclose(result, [[1.0, 0.65, 0.28], [0.82, 0.96, 0.85], [1.0, 0.36, 1.0]])
+
+
+def test_gamma_overlay():
+    """Test that GAMMA overlay works as expected."""
+    result = gamma_overlay(data=RASTERS_DATA, gamma=0.6)
+    assert np.array_equal(
```
nialov, line 42: I see `np.testing.assert_array_equal` often used which does the `assert` within. Might be more suitable for use in `pytest` tests. Might do more fancy reporting of the differences.

```suggestion
    np.testing.assert_array_equal(
```

notebooks/fuzzy_wrapper.ipynb
```diff
@@ -0,0 +1,100 @@
...
+       "'/eis_toolkit'"
+      ]
+     },
+     "execution_count": 32,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "import os\n",
```
nialov, line 21: The notebook seems deprecated? Remove if unused.

nialov (changes requested):
The logic seems solid. The change to `AND` and `OR` logic seems correct in your version now. Might require another look by an expert on these things later.

eis_toolkit/prediction/fuzzy_overlay.py
nmaarnio: I thought about this briefly, but did not do it because the raster data would be then checked three times while calling gamma_overlay. Do you think the min and max checks are really quick even for large raster data? If so, I will change it as you suggested

nmaarnio: The checking could be reduced to two times by omitting input data checking from the gamma_overlay function actually

eis_toolkit/prediction/fuzzy_overlay.py
nmaarnio: Good point. And the check can be simply `if data.min() < 0 or data.max() > 1:`, no need to loop bands.

tests/prediction/fuzzy_test.py
nmaarnio: Added this, and similar for _equivalent_sum_

eis_toolkit/prediction/fuzzy_overlay.py
nialov: True about the performance loss. But I would prioritize in these cases maintainability rather than optimization. You could use memory caching to make sure that same min and max are not calculated multiple times with `functools.cache` (<https://docs.python.org/3/library/functools.html#functools.cache>). I believe it could be used as a decorator for the check function.

nmaarnio: I tried this, but there seemed to be some problem (said _unhashable type: 'numpy.ndarray'_, I did not investigate further). Anyway, with quite a huge array of 80M cells gamma overlay with two checks was performed in 1s, so I do not think this need optimizing at this point.

nialov (approved):
LGTM! I believe this PR could be a good candidate for squashing commits as Johanna's commits probably do not add anything useful to the commit history. Up to you @nmaarnio !

PR #85, opened 2023-01-10
84 improve snap raster function

nmaarnio:
Improved snap raster functionality by the following ways:

- Removed raster band parameter (this information can be extracted directly from the input raster)
- Removed constraint to have input raster within snap raster
- Removed constraint to have as large or larger snap raster pixel size compared to input raster
- Added and modified tests based on the changes

- Additionally, some linting that was missing from extract_values_from_raster.py


lehtonenp:
Hello! I was not aware of these lints as the invoke lint did not warn me about those. My apologies.

PR #86, opened 2023-01-10
Add unique grid combinations

lehtonenp:
This functionality creates a raster of unique combinations using 2 or more raster bands. The bands can be in different files.
For example, if the user would input two rasters with the following values:
[0,1,0,1,1]
[1,1,1,1,0] 
The output combinations would be:
[1,2,1,2,3].
The number 1 would be the first combination, 2 is the second combination, and 3 is the third combination found. While the first and third combinations have the same values but in different rasters, those are still unique combinations.


eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def _unique_combinations(  # type: ignore[no-any-unimported]
+    list_of_rasters: List[rasterio.io.DatasetReader],
+    path_to_output_file: str
+) -> rasterio.io.DatasetReader:
+
+    bands = []
+    for raster in list_of_rasters:
+        bands.append(raster.read())
```
nmaarnio, line 17: I think here you should read the raster band by band and append each band separately, because now you are reading the whole data array with all of its bands straight to the bands list. So each element of the list will be 3D array, even if the raster has only one band.

eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,68 @@
...
+
+    output_band = np.empty_like(bands[0])
+    combinations: Dict[tuple, int] = {}
+    for row, column in itertools.product(range(len(bands[0])), range(len(bands[0][0]))):
+
+        combination = tuple(bands[band][row][column] for band, _ in enumerate(bands))
+
```
nmaarnio, line 24: At the moment this does not work, but after the bands array will actually have bands as elements it should work I think.

eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,68 @@
...
+        if combination not in combinations:
+            combinations[combination] = len(combinations) + 1
+        output_band[row][column] = combinations[combination]
+
+    with rasterio.open(path_to_output_file, 'w', **list_of_rasters[0].meta) as write_raster:
+        write_raster.write(output_band, 1)
+    output_raster = rasterio.open(path_to_output_file)
+    return output_raster
```
nmaarnio, line 32: All the other raster processing functions so far have given "out_image, out_meta" (data array, meta) as the output, and we could follow that convention here too. Writing will be done in a separate function.

eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,68 @@
...
+    Returns:
+        rasterio.io.DatasetReader: Raster with unique combinations
+    """
+    rasters = []
+    for raster in list_of_rasters:
+        rasters.append(raster.read())
+
+    row_length = len(rasters[0][0])
+    column_length = len(rasters[0][0][0])
+
```
nmaarnio, line 57: Just a thought, perhaps it would be more intuitive to access raster.height and raster.width here. But these are correct too.

eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,68 @@
...
+    Args:
+        list_of_rasters (List[rasterio.io.DatasetReader]): Rasters to be used for finding combinations.
+        path_to_output_file (str): The output file location and name for the output.
+
+    Returns:
+        rasterio.io.DatasetReader: Raster with unique combinations
+    """
+    rasters = []
+    for raster in list_of_rasters:
+        rasters.append(raster.read())
```
nmaarnio, line 53: If we read the rasters here already, we could avoid reading them later in the inner function (just thinking if this would be resourcewise easier). But since we want individual bands in the inner function, then this too should be considered already here.

tests/unique_combinations_test.py
```diff
@@ -0,0 +1,37 @@
...
+    assert unique_combinations_raster.meta == raster_1.meta
+    assert unique_combinations_raster.read(1)[0].tolist() == expected_1st_row
+    assert unique_combinations_raster.read(1)[1].tolist() == expected_2nd_row
+
+
+def test_unique_combinations():
+    """Test that invalid parameter values for rasters raises correct exception."""
+    with pytest.raises(InvalidParameterValueException):
+        with rasterio.open(raster_path) as raster:
+            unique_combinations([raster], output_raster_path)
```
nmaarnio, line 37: At the moment this function shadows the first test function since they have the same name. When you rename this one, the first test is executed (and fails, because of the reading rasters instead of raster bands -issue I commented above).

nmaarnio (changes requested):
Hi! I pointed out a few problems as comments. Besides these, I wondered more generally whether this unique_combinations should take rasters as inputs or 2D arrays (basically raster bands). If this functionality is used only for rasters, it makes sense to take rasters as inputs, otherwise I would opt for a more general array function. Secondly, I'm thinking if this function is supposed to take only rasters with same grid size and location (so, unified/co-registered rasters), or is it okay if the rasters do not actually align? Currently this is not tested in any way at least. Maybe @jtorppa or @chudasama-bijal have an opinion on these too?

eis_toolkit/raster_processing/unique_combinations.py
lehtonenp: Thank you for the info. I was unaware of this, albeit the fact that every other function gives out those outputs. Maybe this is something that should be mentioned to @jtorppa, too.

nmaarnio:
Looks good to me now! I noticed however that the linter wants to change the single quotation marks to double ones on lines 44, 45 and 46. Also it wants to change the order of some imports and add spaces around "+" on line 49. Did you forget to run the linter, or is there some differences how it behaves in different development environments I wonder? If the latter, I need to investigate this matter soon

lehtonenp:
> I noticed however that the linter wants to change the single quotation marks to double ones on lines 44, 45 and 46. Also it wants to change the order of some imports and add spaces around "+" on line 49...

I ran the linter and got no errors where you pointed out. Could you then tell me the proper order of the imports?

nmaarnio:
Okay, good to know. This is the order of imports for me:
import itertools
from typing import Dict, List, Tuple

import numpy as np
import rasterio

from eis_toolkit.exceptions import InvalidParameterValueException

tests/unique_combinations_test.py
```diff
@@ -0,0 +1,37 @@
...
+expected_1st_row = [1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0, 12.0, 13.0, 14.0, 15.0, 16.0, 17.0, 18.0,
+19.0, 20.0, 21.0, 22.0, 23.0, 24.0, 25.0, 26.0, 27.0, 28.0, 29.0, 30.0, 31.0, 32.0, 33.0, 34.0, 35.0, 36.0,
+37.0, 38.0, 39.0, 40.0, 41.0, 42.0, 43.0, 44.0, 45.0, 46.0]
+expected_2nd_row = [47.0, 48.0, 49.0, 50.0, 51.0, 52.0, 53.0, 54.0, 55.0, 56.0, 57.0, 58.0, 59.0, 60.0, 61.0, 62.0,
+63.0, 64.0, 65.0, 66.0, 67.0, 68.0, 69.0, 70.0, 71.0, 72.0, 73.0, 74.0, 75.0, 76.0, 77.0, 78.0, 79.0, 80.0, 81.0, 82.0,
+83.0, 84.0, 85.0, 86.0, 87.0, 88.0, 82.0, 89.0, 90.0, 91.0]
+
+def test_unique_combinations():
+    """Test unique combinations creates expected output."""
+    raster_1 = rasterio.open(raster_path)
```
nialov, line 23: You can do multiple context managers with single indent:

```suggestion
    with rasterio.open(raster_path) as raster_1, rasterio.open(raster_path) as raster_2:
```

eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    rasters = []
+    out_meta = raster_list[0].meta
+    out_meta['count'] = 1
+    height = raster_list[0].meta['height']
+    width = raster_list[0].meta['width']
+
+    for raster in raster_list:
+        for band in range(1, raster.count+1):
+            rasters.append(raster.read(band))
```
nialov, line 50: It is quite easy to run out of memory when reading rasters fully into memory as `numpy` arrays. But as long as the rasters are small, this is not a concern. Just something to keep in mind.

lehtonenp: Thanks for the info. Is there another way to read the rasters and use them in the function?

nialov: @jtlait uses windowed reads here already in `eis_toolkit`: https://github.com/GispoCoding/eis_toolkit/blob/77fba1fc64755f898d7849bb79c7b86df6b3211d/eis_toolkit/raster_processing/windowing.py#L11

Most relevant lines:

~~~python
    window = Window(
        col_off=top_left_col,
        row_off=top_left_row,
        width=width,
        height=height,
    )

    out_image = raster.read(
        boundless=True,
        window=window,
        fill_value=-9999,
    )
~~~

The basic principle is to read only a "window" of the raster into an array. I could not readily say how easy it is to implement for the functionality you are doing here. It is more difficult to work with windows of data and then in the end make sure you have correctly aggregated the results from all the windows.

lehtonenp: Thank you for the tip. I'll keep this in mind as I am not sure how to implement this, yet.

nmaarnio: Returning to this PR and conversation, I think we should, for now, go with the initial implementation of reading the raster fully into memory. There are other functions already present in the EIS Toolkit that read rasters similarly and don't take these extra measures for memory management, so if we decide to take them (maybe find out in the testing that it is necessary), then we can update multiple problematic functions at once.

lehtonenp: I have now made some optimizations to the code. With the `data/remote/workflow_demo/IOCG_AEM_Inph_.tif` file I got the execution from ~4 seconds to ~2 seconds. However, there is a triple nested loop in the new implementation that could cause slowness with bigger datasets.

eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,75 @@
...
+            for column in range(band.shape[1]):
+                raster_value = band[row, column]
+                if raster_value not in unique_combinations:
+                    unique_combinations[raster_value] = combination
+                    combination += 1
+                unique_indices[row, column] = unique_combinations[raster_value]
+
+    return unique_indices
+
+
```
nmaarnio, line 28: Please add @beartype decorator here. We picked up this handy package to do input validation some time ago, you can check some existing functions for reference.

eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,75 @@
...
+        raise InvalidParameterValueException("Expected all raster files to have the same pixel height")
+
+    if not all(raster.meta["crs"] == out_meta["crs"] for raster in raster_list):
+        raise InvalidParameterValueException("Expected all raster files to have the same coordinate system")
+
+    if not all(len(band) == height for band in bands):
+        raise InvalidParameterValueException("Expected all raster files to have the same height")
+
+    if not all(len(band[0]) == width for band in bands):
+        raise InvalidParameterValueException("Expected all raster files to have the same width")
```
nmaarnio, line 72: You could replace these checks with a premade check from the `eis_toolkit.checks.raster` module. There's one called `check_raster_grids` that combines everything you are doing here. The only downside in using it that your exception message would be generic and not identify what is wrong with the grids. But I think it is not too bad for now and perhaps we should make the individual checks used in `check_raster_grids` to raise exceptions automatically in the future

tests/raster_processing/unique_combinations_test.py
nmaarnio, line 1: Right now all the combinations checked are different, i.e. there is no combination repetition as there was in the short description of this PR. I would prefer a similar example/test here, so perhaps something like two 4x4 rasters would suffice. We could make these and add these to test data / write the test data as raster to memory during test execution.

eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,75 @@
...
+    combination = 1
+
+    for band in bands:
+        for row in range(band.shape[0]):
+            for column in range(band.shape[1]):
+                raster_value = band[row, column]
+                if raster_value not in unique_combinations:
+                    unique_combinations[raster_value] = combination
+                    combination += 1
+                unique_indices[row, column] = unique_combinations[raster_value]
```
nmaarnio, line 24: I'm having a little trouble trying to verify this works as intended. Does this consider combinations across all bands right now? It seems it might only check if individual value was encountered previously. If this is the case, you could leave looping of the bands out and inside the row-col loop do something like this:

`combination = tuple(band[row, col] for band in bands)`
`unique_indices[row, col] = unique_combinations[combination]`

Also while looking at this PR I happened to find this neat tool you could maybe use here: `defaultdict`.

nmaarnio (commented):
Hi, I took a look at this and wrote a few comments.

tests/raster_processing/unique_combinations_test.py
lehtonenp: I got similar feedback from @em-t, as well. I'll dig into this. As the inner function takes in np.ndarrays I can test the functionality with that.

eis_toolkit/raster_processing/unique_combinations.py
lehtonenp: @nmaarnio I have now used numpy to get the unique combinations. It is faster with a 10k x 10k grid compared to the old one (95 sec to 85 sec). I apologize for the bad naming of the commits. I made commits called 'Use numpy to find unique combinations without nested loops'. Somehow, with my lack of understanding of pre-commit, the commits got lost in the way.

lehtonenp:
@nmaarnio I have now used numpy to get the unique combinations. It is faster with a 10k x 10k grid compared to the old one (95 sec to 85 sec). I apologize for the bad naming of the commits. I made commits called 'Use numpy to find unique combinations without nested loops'. Somehow, with my lack of understanding of pre-commit, the commits got lost in the way.

lehtonenp:
> @lehtonenp is this ready for a new review, or are you still planning to make modifications?

@nmaarnio this is now ready for review. The function should work as intended but the combination number is not in running order.

eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,58 @@
...
+
+    All bands in all rasters are used for analysis.
+    The first band of the first raster is used for reference when making the output.
+
+    Args:
+        raster_list (List[rasterio.io.DatasetReader]): Rasters to be used for finding combinations.
+
+    Returns:
+        out_image (numpy.ndarray): Combinations of rasters.
+        out_meta (dict): The metadata of the first raster in raster_list.
```
nmaarnio, line 41: We have decided to omit the parameter and return value types from the docstings.

eis_toolkit/raster_processing/unique_combinations.py
```diff
@@ -0,0 +1,58 @@
...
+    unique_combinations = inverse_indices.reshape(bands[0].shape)
+
+    unique_combinations += 1
+
+    return unique_combinations
+
+
+@beartype
+def unique_combinations(  # type: ignore[no-any-unimported]
+    raster_list: List[rasterio.io.DatasetReader],
```
nmaarnio, line 29: You could change the `List` annotation into a `Sequence`, just so that beartype won't get mad if a user will give a tuple. Also change this in the inner function

nmaarnio (commented):
Just a couple nitpicky comments regarding docstring convention and typing, otherwise looks good!

nmaarnio (approved):

nmaarnio:
LGTM, merging!

PR #91, opened 2023-01-24
Initial functionality for csv to geopandas conversion

Mtk112:
Added functionality for csv to geopandas dataframe conversion. Fixed testing_clip.ipynb.


eis_toolkit/conversions/csv_to_geopandas.py
```diff
@@ -0,0 +1,98 @@
...
+            try:
+                geodataframe = geopandas.GeoDataFrame(
+                    df, crs=target_EPSG, geometry=geopandas.points_from_xy(df[indexes[0]], df[indexes[1]]))
+                return geodataframe
+            except:
+                raise InvalidParameterValueException
+
+
+def csv_to_geopandas(  # type: ignore[no-any-unimported]
+    csv: str,
```
nialov, line 74: Is there a reason not to use `pathlib.Path` (`from pathlib import Path`) for paths here rather than passing the path as as a `str`? I believe it is already used elsewhere in `eis_toolkit`. It helps to avoid errors later with path separator differences between Windows and linux-based operating systems. Also less ambiguous as a `str` type can represent any strings rather than only file paths.

```suggestion
    csv: Path,
```

eis_toolkit/conversions/csv_to_geopandas.py
```diff
@@ -0,0 +1,98 @@
...
+                                    InvalidColumnIndexException)
+
+
+def _csv_to_geopandas(  # type: ignore[no-any-unimported]
+    csv: str,
+    indexes: List[int],
+    target_EPSG: int,
+) -> geopandas.GeoDataFrame:
+
+    with open(csv, 'r') as f:
```
nialov, line 18: If `pathlib.Path` is used, it exposes an alternative interface for opening a text file:

```suggestion
    with csv.open(mode="r") as f:
```

although the original method should work as well. Just an alternative..

eis_toolkit/conversions/csv_to_geopandas.py
```diff
@@ -0,0 +1,98 @@
...
+def csv_to_geopandas(  # type: ignore[no-any-unimported]
+    csv: str,
+    indexes: List[int],
+    target_EPSG: int,
+) -> geopandas.GeoDataFrame:
+    """
+    Convert CSV file to geopandas DataFrame.
+
+    Usage of single index expects valid WKT geometry.
+    Usage of two indexes expects the X and Y columns of POINT feature(s).
```
nialov, line 82: Should explicitly state that X-coordinate must be the first index and Y-coordinate the second.

tests/csv_to_geopandas_test.py
```diff
@@ -0,0 +1,53 @@
...
+    with pytest.raises(InvalidWktFormatException):
+        indexes = [3]
+        target_EPSG = 4326
+        csv_to_geopandas(csv_path, indexes, target_EPSG)
+
+def test_csv_to_geopandas_points():
+    """Test csv with point features to geopandas conversion using lattitude and longitude."""
+    indexes = [5,6]
+    target_EPSG = 4326
+    csv_to_geopandas(csv_path, indexes, target_EPSG)
```
nialov, line 39: Could do some additional sanity checks

```suggestion
    gdf = csv_to_geopandas(csv_path, indexes, target_EPSG)
    assert isinstance(gdf, gpd.GeoDataframe)
    assert gdf.crs.to_epsg() == target_epsg
```

nialov (commented):
Personally, I would prefer using column names rather than column indexes if at all possible. I believe it would be more convenient for an user anyway.

Otherwise some general suggestions.

eis_toolkit/conversions/csv_to_geopandas.py
Mtk112: Thank you for the comments. The reason I opted for using indexes rather than column names is so that the function could handle .csv files without headers. As for the general suggestions, I am implementing all of the suggested changes.

tests/csv_to_geopandas_test.py
Mtk112: Thank you for the suggestion, I will implement these changes.

tests/csv_to_geopandas_test.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def test_csv_to_geopandas_using_wkt_invalid_parameter_value():
+    """Test that index out of range raises correct exception."""
+    with pytest.raises(InvalidColumnIndexException):
+        indexes = [8]
+        target_EPSG = 4326
+        gdf = csv_to_geopandas(csv_path, indexes, target_EPSG)
+        assert isinstance(gdf, gpd.GeoDataFrame)
+        assert gdf.crs.to_epsg() == target_EPSG
```
nmaarnio, line 32: For all the tests that are supposed to raise an Exception, you don't need the assert statements, only the call to the csv_to_geopandas

nmaarnio (commented):
There was only one thing about tests that I commented, otherwise I think this is good to go. We can iterate the implementation later, if we want to have column names as inputs or restrict/broaden the uses of this function.

nmaarnio (approved):
Merging

PR #92, opened 2023-01-25
Create PCA.py

dipak6697:
This is the PCA implementation

### DIFF
--- a/tests/PCA.py
+++ b/tests/PCA.py
@@ -1,0 +1,60 @@
+import pandas as pd
+import numpy as np
+from sklearn.decomposition import PCA
+from sklearn.preprocessing import StandardScaler
+import matplotlib
+import matplotlib.pyplot as plt
+
+
+# Reading the data
+
+num_components = 2
+#importing 1112X9 data frame
+path1 = "1112x9.csv"
+#importing 17X9 data frame
+path2 = "17x9.csv"
+df1 = pd.read_csv(path1)
+df2 = pd.read_csv(path2)
+
+#defining the function for PCA
+
+def MyPCA(X):
+    # standardize the Data
+    features1 = X.columns
+
+    # Separating out the features
+    x1 = X.loc[:, features1].values
+
+    # Standardizing the features
+    x1 = StandardScaler().fit_transform(x1)
+
+    # Calculating the PCA
+    pca = PCA(n_components=num_components)
+    principalComponents1 = pca.fit_transform(x1)
+    dataframe = pd.DataFrame(data=principalComponents1, columns=['principal_component_1', 'principal_component_2'])
+
+    return dataframe
+
+
+# Defining the function for plotting the PCA
+
+def PCA_plot(X1, X2):
+    #plotting the graph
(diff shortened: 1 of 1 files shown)


tests/PCA.py
```diff
@@ -0,0 +1,60 @@
...
+# Reading the data
+
+num_components = 2
+#importing 1112X9 data frame
+path1 = "1112x9.csv"
+#importing 17X9 data frame
+path2 = "17x9.csv"
+df1 = pd.read_csv(path1)
+df2 = pd.read_csv(path2)
+
```
nmaarnio, line 18: Please move all this to a separate PCA_test.py file, where the PCA function is run on actual data. You can check how test files for other functions look like for reference. Please also provide either the csv-file or hard code the data directly into the dataframe.

tests/PCA.py
```diff
@@ -0,0 +1,60 @@
...
+#defining the function for PCA
+
+def MyPCA(X):
```
nmaarnio, line 21: You can remove the "My" from the function name. Also, please have a more descriptive name for the function argument and use type annotations.

```suggestion
def PCA(dataframe: pd.DataFrame) -> pd.DataFrame:
```

tests/PCA.py
```diff
@@ -0,0 +1,60 @@
...
+def PCA_plot(X1, X2):
+    #plotting the graph
+    plt.scatter(X2.principal_component_1, X2.principal_component_2,marker='.', color='blue', s=200, alpha = .2)
+    plt.scatter(X1.principal_component_1, X1.principal_component_2,marker='*', color='red', s=500)
+    matplotlib.rcParams.update({'font.size': 22})
+    plt.title("Principal components of all points")
+    plt.xlabel('Principal component 1')
+    plt.ylabel('Principal component 2')
+
+    plt.show()
```
nmaarnio, line 50: We have a plan to implement plotting functions separately and more generally, so that multiple functions can call one plotting function with different arguments. However, for now you can have the plotting here, and we can remove it later.

tests/PCA.py
```diff
@@ -0,0 +1,60 @@
...
+    
+    
+# Calling the function for PCA
+
+principal_df1 = MyPCA(X=df1)
+principal_df2 = MyPCA(X=df2)
+
+#calling the function for scatter plot
+
+PCA_plot(principal_df2,principal_df1)
```
nmaarnio, line 60: This would also belong in the test file.

nmaarnio (changes requested):
Thanks for showing this function for review so quickly! I wrote a few comments on specific places on the code, and here are some general notes:
- Please put the PCA.py under exploratory analysis -folder, and put a PCA test file in the tests folder
- Try to copy the form of the already implemented functions
- Remember to run the "pytest" and "invoke lint" commands when you have updated the code

tests/PCA.py
dipak6697: done now!

tests/PCA.py
dipak6697: this is done

dipak6697 (commented):
I did the requested modifications.

nmaarnio:
A new PR was opened regarding the same addition

PR #95, opened 2023-02-14
Pull request pca

dipak6697:
We remodified  the code as per your comment

### DIFF
--- a/tests/PCA.py
+++ b/tests/PCA.py
@@ -5,56 +5,36 @@ from sklearn.preprocessing import StandardScaler
 import matplotlib
 import matplotlib.pyplot as plt
 
-
-# Reading the data
-
-num_components = 2
-#importing 1112X9 data frame
-path1 = "1112x9.csv"
-#importing 17X9 data frame
-path2 = "17x9.csv"
-df1 = pd.read_csv(path1)
-df2 = pd.read_csv(path2)
-
 #defining the function for PCA
+def Compute_PCA(data_to_compute, num_components, show_results=False):
+    """THIS FUNCTOIN TAKES DATAFRAME AS INPUT, CALCULATES THE PRINCIPLE COMPONENTS AND RETURS A DATAFRAME CONTAINING THE PRINCIPAL COMPONETS INSIDE IT"""
+    try:
+        # Separating out the features
+        data = data_to_compute.loc[:, data_to_compute.columns].values
 
-def MyPCA(X):
-    # standardize the Data
-    features1 = X.columns
+        # Standardizing the features
+        data = StandardScaler().fit_transform(data)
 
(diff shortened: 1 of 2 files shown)


tests/PCA.py
```diff
@@ -0,0 +1,40 @@
+import pandas as pd
+import numpy as np
+from sklearn.decomposition import PCA
+from sklearn.preprocessing import StandardScaler
+import matplotlib
+import matplotlib.pyplot as plt
+
+#defining the function for PCA
+def Compute_PCA(data_to_compute, num_components, show_results=False):
+    """THIS FUNCTOIN TAKES DATAFRAME AS INPUT, CALCULATES THE PRINCIPLE COMPONENTS AND RETURS A DATAFRAME CONTAINING THE PRINCIPAL COMPONETS INSIDE IT"""
```
nmaarnio, line 10: Please modify the docstring so that it looks like the docstrings in our other toolkit functions. Take a look at for example the clipping function. I suggest you copy-paste the docstring of clip for example, and modify the contents according your function.

nmaarnio, line 10: Please modify the docstring so that it looks like the docstrings in our other toolkit functions. Take a look at for example the clipping function. I suggest you copy-paste the docstring of clip for example, and modify the contents according your function.

tests/PCA.py
```diff
@@ -0,0 +1,40 @@
...
+    try:
```
nmaarnio, line 11: We want to avoid using generic try - except -blocks. If there are some specific requirements for the data, write those as checks (using if expressions for example) and raise a custom expression if something is wrong. So please remove the try and except and if needed, add some other code to handle error situations. Again I suggest taking look at other toolkit functions.

tests/PCA.py
```diff
@@ -0,0 +1,40 @@
...
+        columns_name_holder = []
+        for i in range(num_components):
+            columns_name_holder.append(f'principal_component_{i+1}')
+
+        # create a data frame with ocs inside 
+        dataframe = pd.DataFrame(data=principalComponent, columns=columns_name_holder)
+
+        # if show the res is true print it on screen 
+        if(show_results):
+            print(dataframe)
```
nmaarnio, line 31: You could remove this for now. I think we don't need to print the dataframe directly in the toolkit function when we know this function is working.

tests/PCA_test.py
```diff
@@ -0,0 +1,26 @@
+import pandas as pd 
+import numpy as np
+from PCA import *
```
nmaarnio, line 3: Import specific classes/functions from our toolkit files when possible
```suggestion
from PCA import Compute_PCA
```

nmaarnio:
A new PR was created for the same feature

PR #97, opened 2023-02-22
Implement a draft command-line interface for `eis_toolkit`

nialov:
Added the basic draft of the interface and implemented an entrypoint for
the `eis_toolkit.raster_processing.clipping.clip_raster` function as
`cli-clip-raster`. Naming and many other details are just as examples,
they can be changed as wanted.

Also added the needed boilerplate to make the command-line interface
installable with `pip` and `poetry` in `pyproject.toml`. Furthermore,
for local development and testing primarily, the command-line can be
called with `python -m eis_toolkit`. Use the `--help` option to show the
help. E.g.

~~~bash
poetry run python -m eis_toolkit --help
~~~

Output currently:

~~~
Usage: python -m eis_toolkit [OPTIONS] COMMAND [ARGS]...

  Click group to nest subcommands under one interface.

Options:
  --help  Show this message and exit.

Commands:
  cli-clip-raster  Clip a raster at ``raster_path`` with geometries in...
~~~

### DIFF
--- a/eis_toolkit/__main__.py
+++ b/eis_toolkit/__main__.py
@@ -1,0 +1,4 @@
+from eis_toolkit.cli import cli
+
+if __name__ == "__main__":
+    cli()
--- a/eis_toolkit/cli.py
+++ b/eis_toolkit/cli.py
@@ -1,9 +1,9 @@
+"""Command-line interface for eis_toolkit."""
 import json
 from pathlib import Path
 
 import click
 import geopandas as gpd
-import jsonschema
 import rasterio
 
 from eis_toolkit.raster_processing.clipping import clip_raster
@@ -62,26 +62,28 @@ def cli_clip_raster_config(config_file: str):
 
     # I added an example on how to validate the configuration passed from a config file.
     # It uses the jsonschema library, which was, like click already a dependency
-    schema = {
-        "type": "object",
-        "properties": {
-            "raster_file": {
-                "type": "string",
-                "minLength": 1,
-            },
-            "geodataframe_file": {
-                "type": "string",
-                "minLength": 1,
-            },
-            "output_raster_file": {
-                "type": "string",
-                "minLength": 1,
-            },
-        },
-        "required": ["raster_file", "geodataframe_file", "output_raster_file"],
-    }
+    # in poetry.lock but not in requirements.txt so I commented this out to avoid
+    # ci errors.
+    # schema = {
+    #     "type": "object",
+    #     "properties": {
+    #         "raster_file": {
+    #             "type": "string",
+    #             "minLength": 1,
+    #         },
+    #         "geodataframe_file": {
+    #             "type": "string",
+    #             "minLength": 1,
+    #         },
+    #         "output_raster_file": {
+    #             "type": "string",
+    #             "minLength": 1,
+    #         },
+    #     },
+    #     "required": ["raster_file", "geodataframe_file", "output_raster_file"],
+    # }
 
-    jsonschema.validate(instance=loaded_config, schema=schema)
+    # jsonschema.validate(instance=loaded_config, schema=schema)
 
     click.echo(loaded_config)
 
--- a/eis_toolkit/cli.py
+++ b/eis_toolkit/cli.py
@@ -1,0 +1,36 @@
+from pathlib import Path
+
+import click
+import geopandas as gpd
+import rasterio
+
+from eis_toolkit.raster_processing.clipping import clip_raster
+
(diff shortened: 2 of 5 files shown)


nialov:
From here I can add some comments to explain how `click` works and also to show how to add `pytest` testing for the command-line interface. I will also draft an interface that uses a `json` configuration file for getting the needed arguments for a function call.

nialov:
Added the `json` config file example implementation:

~~~bash
poetry run python -m eis_toolkit cli-clip-raster-config tests/data/remote/clip_raster_cli_test_config.json
~~~

Output:
~~~
{'raster_file': 'tests/data/remote/small_raster.tif', 'geodataframe_file': 'tests/data/remote/small_area.shp', 'output_raster_file': 'tests/data/local/test.tif'}
Reading geodataframe at tests/data/remote/small_area.shp.
Opening raster at tests/data/remote/small_raster.tif.
Writing raster to tests/data/local/test.tif.
~~~

nialov:
Added `pytest` tests for the current interfaces using `click.testing.CliRunner` in `tests/cli_test.py`.

nmaarnio:
Thanks for doing this draft. I am still thinking what option would be the best. If I read this correctly, Click does not provide a solution to the problem I was most worried about, which is that we need to define a specific entry point for each function of the toolkit. However, at this point I have a few ideas for structuring the functions that might help, but I haven't yet had much time to use on this interface design. About Click still, if we assumed no one would use the CLI itself, do you think using Click would still be beneficial and worth the extra code? If we anyway need individual entry points for toolkit functions, I'm thinking the communication between toolkit and plugin would be so "defined" that using sys.argv would suffice

nialov:
> Thanks for doing this draft. I am still thinking what option would be the best. If I read this correctly, Click does not provide a solution to the problem I was most worried about, which is that we need to define a specific entry point for each function of the toolkit. However, at this point I have a few ideas for structuring the functions that might help, but I haven't yet had much time to use on this interface design. About Click still, if we assumed no one would use the CLI itself, do you think using Click would still be beneficial and worth the extra code?

This can be done programmatically. The approach I have used is just the conventional one of decorating explicitly declared functions. No reason why you could not do something like this:

~~~python

@click.group()
def cli():
    """Click group to nest subcommands under one interface."""
    pass

def add_to_cli_group(function, cli=cli):
    ...

function_names = [ "clip_raster", "snap_with_raster" ]

for f in function_names:
    add_to_cli_group(function=f)
~~~

However, I have not personally done this with `click`. It will be a bit opaque/complex as it does not follow the typical convention. The advantage is the automatic generation of help documentation for the command-line with `--help`:

~~~
Usage: python -m eis_toolkit [OPTIONS] COMMAND [ARGS]...

  Click group to nest subcommands under one interface.

Options:
  --help  Show this message and exit.

Commands:
  cli-clip-raster  Clip a raster at ``raster_path`` with geometries in...
~~~

>  If we anyway need individual entry points for toolkit functions, I'm thinking the communication between toolkit and plugin would be so "defined" that using sys.argv would suffice

Sufficient? Probably yes. I like to warn about path handling in these cases but I could not say for certain if e.g. calls through `subprocess` from the GUI side will have issues with Windows paths with spaces for example. `click` should handle these cases already or at least fail them gracefully. Testing with the battle-hardened `click.testing` utilities is also an advantage of using `click` (although there is not that much to test if all entrypoints only take a config file).

# Other thoughts

There are some things to keep in mind with a command-line interface such as this. Each call through `subprocess` to the `eis_toolkit` will require re-initialization of the Python environment which can take some time due to the toolkit using machine learning libraries (e.g. `import tensorflow` will be itself take some time from my experience). So there will be some slowdown the more calls there are through the command-line interface. This could be avoided if the workflows (=multiple functions calls) were implemented in `eis_toolkit` itself as @nmaarnio had drafted previously. This could also be avoided if the toolkit was implemented as an `API` and it would constantly run in the background but that might be a bit too complex and error-prone. Not saying this invalidates using entrypoints for each function, as I have drafted here, but just something to keep in mind. If hundreds of calls are required for single workflows through the interface then the slowdown might be too much.

nmaarnio:
I will familiarize myself with Click in the near future and draft some skeleton of an interface myself too. Anyway, to clarify this task, I can be the main developer implementing it since I will be coding the QGIS plugin too, but I will continue to ask your opinion and review on this feature in the future too, if you have bandwidth @nialov

nmaarnio:
Closing this draft PR since drafting the CLI has ended and the actual implementation (using Typer) has been started.

PR #98, opened 2023-02-24
22 add transformation functions

msmiyels:
This push req. contains all the listed transformation functions and some additions:
- binarize
- winsorize by absolute values
- winsorize by percentile values
- logarithmic transformation
- sigmoid transformation
- linear transformation with z-score normalization
- linear transformation with min-max-scaling

Beside the core functions, some separate input-parameter checks were added to the check-section of the toolkit and a also a utils file, containing some important helper functions, comes along with the core functions.

Functionality is build the way, that there is an "outside" function containing the docstring etc., which raises a second function, either for raster or dataframe processing with different arguments. However, both raster and dataframe functions call the same core function for transformation calculations.

Additionally, a notebook is provited, containing explanations, examples and also examples raising different errors (expecially wrong inputs). Also, a new small raster file was added to the remote directory, which was used to test the functionalities.

For all transformation functions, a single input parameter-value can used to transform every band or column (everything with same parameters). However, it is also possible to specify different parameters for every band or column. How this works in detail is shown in the supplied notebook.

All functions return the transformed array and important input settings with calculated statistical values during transformation for every band/column. For raster data, the metadata will be returned as well. Since dataframes do not have this kind of metadata, a dictionary of the "transformable" (i.e. numeric) columns, geometry columns etc. will be returned instead. This is quite helpful for later use to convert the array back into a dataframe.

To convert a resulting array back to a dataframe, it needs to be transposed (array.T). However, if the returned array has already to be in that shape, we can change that quickly later on.


nmaarnio (commented):
Sorry for the delay in reviewing! Here are a few things I would suggest to increase code reuse and make the logic simpler:
- Let's remove the "method" parameter from the functions, I think this functionality should not be part of transformation functions. Same as with different nodata values for each band, I think this kind of raster handling should be done separately. We could either give the user to define one nodata value, or not even give this option (although it might be needed for vector data).
- All the "valids" lists can be left out if we implement the argument type check utility function, and then check the parameter values as necessary. If some checks are common, let's make a reusable check out of them (perhaps checks like "check bands valid", "check columns valid", "check columns numeric" as I have drafted)
- I am somewhat torn trying to decide if the "out_settings" is needed, but leaning towards leaving it out. Do you think it is useful? Especially when used from QGIS, I do not think these would ever be used, and in scripts one could pretty easily fetch this information from other parts of the code too, probably.
- We could increase code reuse with nodata handling too. I have drafted a decorator we could apply, I will soon show in a draft PR linked to utility functions issue.

Other stuff:
- As a general comment, I would suggest rechecking the need for the utility functions and checks I did not comment but that you have implemented, after you have modified the code. There might be something more to be streamlined.
-  I made some comments within the code pointing out minor things. In addition I also put there some untested draft code I created in the process of trying to simplify and clean the logic.
- I would like to have actual assertions in the tests, showing that the output is as intended. Now the tests only test exceptions.
- I will try to work on the first set of utility functions this week. It would be good if you could use them in these transformation functions.

Thanks for such a comprehensive implementation, and again sorry for the delay.

eis_toolkit/transformations/minmax_scaling.py
```diff
@@ -0,0 +1,141 @@
...
+import pandas as pd
+import geopandas as gpd
+import rasterio
+from typing import Optional, Tuple, Union, List, Literal
+
+from eis_toolkit.exceptions import InvalidParameterValueException, InvalidInputDataException
+from eis_toolkit.transformations import utils
+from eis_toolkit.checks import parameter
+
+
```
nmaarnio, line 11: While trying to understand the code and come up with improvement suggestions, I ended up drafting an alternative implementation and used minmax as an example. Note that this is not tested code! Also, I made it mainly for you to get ideas, no need to take this suggestion super strictly.
```suggestion
@check_arguments
def minmax_scale(
    in_data: pd.DataFrame | gpd.GeoDataFrame | rasterio.io.DatasetReader,
    selection: Optional[List[int | str]] = None,
    new_ranges: List[Tuple[np.number, np.number]] = [(0, 1)]
) -> np.ndarray:
    
    if selection is not None and len(new_ranges) != 1 and len(selection) != len(new_ranges):
        raise InvalidParameterValueException("Specify scaling range for each selection or only one range")

    if isinstance(in_data, rasterio.DatasetReader):
        out_array = _minmax_scale_raster(in_data, selection, new_ranges)

    elif isinstance(in_data, pd.DataFrame | gpd.GeoDataFrame):
        out_array = _minmax_scale_vector(in_data, selection, new_ranges)
        
    return out_array


def _minmax_scale_raster(
    raster: rasterio.io.DatasetReader,
    selection: Optional[List[int]],
    new_ranges: List[Tuple[np.number, np.number]]
) -> np.ndarray:
    
    if selection is None:
        selection = list(range(1, raster.count + 1))
    else:
        check_bands_valid(raster, selection)

    out_array = raster.read()

    for band, new_range in expand_and_zip(selection, new_ranges):
        in_data = raster.read(band)
        nodata_values = raster.meta["nodata"]
        out_data = _minmax_scaling_core(in_data, nodata_values, new_range)
        out_array[band] = out_data

    return out_array


def _minmax_scale_vector(
    vector_data: pd.DataFrame | gpd.GeoDataFrame,
    selection: Optional[List[str]],
    new_ranges: List[Tuple[np.number, np.number]]
) -> np.ndarray:

    if selection is None:
        selection = vector_data.select_dtypes(include=[np.number]).columns.to_list()
        check_columns_valid(vector_data, selection)
    else:
        check_columns_valid(vector_data, selection)
        check_columns_numeric(vector_data, selection)

    out_data = vector_data.copy()

    for column, new_range in expand_and_zip(selection, new_ranges):
        series_as_array = out_data[column].to_numpy()
        nodata_values = "todo"
        out_data = _minmax_scaling_core(series_as_array, nodata_values, new_range)
        out_data[column] = out_data

    return out_data


@handle_nodata_as_nan
def _minmax_scaling_core(
    data: np.ndarray,
    new_range: Tuple[np.number, np.number],
) -> np.ndarray:
    
    min = np.nanmin(data)
    max = np.nanmax(data)
    scaled_min = new_range[0]
    scaled_max = new_range[1]
    
    scaler = (data - min) / (max - min)
    out_data = (scaler * (scaled_max - scaled_min)) + scaled_min
    
    return out_data
```

nialov:
Just to let you know, I will take a look at this PR as well by the end of this week. I will try to provide some concrete suggestions similar to Niko and check his recommendations.

eis_toolkit/transformations/binarize.py
```diff
@@ -0,0 +1,189 @@
...
+) -> np.ndarray:
+    
+    out_array = data_array
+    out_array = utils.replace_nan(data_array=out_array, nodata_value=nodata_value, set_nan=True)
+    out_array = np.ma.array(out_array, mask=np.isnan(out_array))
+    
+    out_array[out_array <= threshold] = 0
+    out_array[out_array > threshold] = 1
+    
+    out_array = out_array.data
```
nialov, line 26: These lines provide the main functionality of binarization. The amount of boilerplate related to just handling multiple bands, geodataframes, multiple columns fill the rest of the file. The boilerplate needs to be reduced.

nialov:
## Handling multiple bands

Do any of the functions actually do analysis based on multiple bands or do they
all take a single band as an input, process it, then move to the next band with
no information being used between the bands? If that is the case, the functions
should be simplified to only work on a single band and the calls to use the
function with multiple bands of a raster should be done upstream. If the bands
are used together, then of course multiple bands need to be available, but this
is not the case for instance for `binarize` where single bands are
binarized individually. Handling multiple bands in each function adds much
complexity. Furthermore, each function reads all the bands of the raster into
memory simultaneously. This should be avoided unless they are actually needed
simultaneously as you may run into memory issues very fast.
        
## Handling both GeoDataFrames and raster

This is a general issue that causes much unneeded complexity. There is no
reason to handle them separately in each toolkit module. Instead, we should
decide on a either or to be used as the input in all raster processing
modules/functions. Rasters from `rasterio` is my preferred choice.
GeoDataFrames, which can be assumed to be point data on a fixed grid, can be
converted to rasters via a rasterization function.

Consequently, if this seems reasonable to others as well (@nmaarnio), I would
focus all refactoring efforts in this pull request to the functions handling
raster data or even just the `_core` functions (See below.)

## Core functionality is hidden beneath the boilerplate

The actual core functionality is currently well separated with the `_core`
function naming suffixes. However, everything else in the files seems related
to handling all bands and GeoDataFrames simultaneously. Same handling with
small differences is implemented multiple times in the different files.
Functions with `_table` and `_raster` suffixes usually do the same thing across
all Python files, all related to just handling the different input data types.

This "boilerplate" functionality, as I call it, should be separated outside
this pull request, but the code done here can simply be copied there and
should not be wasted.

If this seems reasonable to others as well (@nmaarnio), I would focus
refactoring efforts to i) separating the general utilities from this pull
request to the general utilities that are being drafted by Niko. Furthermore,
ii) I would add separate tests for all the `_core` functions to make sure they
perform as expected as they are the critical part of this pull request. They
should be tested with assertions and checks that a previously known, expected,
result is returned from the functions.

## Comments by Niko

> I am somewhat torn trying to decide if the "out_settings" is needed, but leaning towards leaving it out. Do you think it is useful? Especially when used from QGIS, I do not think these would ever be used, and in scripts one could pretty easily fetch this information from other parts of the code too, probably.

Any relevant default values set in `eis_toolkit` should be adjustable from the
QGIS interface. So the user should already know before running a function
there, what values he is using in the function. In this case I do not see the
need for `out_settings`. However, if defaults are used in code that are not
adjustable by the user, these should be somehow reported (probably not a good
practice). This is more of a general issue than anything related to
`transformations` module.

> I would like to have actual assertions in the tests, showing that the output is as intended. Now the tests only test exceptions.

Currently, very little is actually tested as there is no checking if the
outputs are even of the types that are expected. I would start by only adding
test cases for the `_core` functions where you check that the outputs are of
the right type and that they fulfill some criteria, even checking that the
array values are exactly the same as some expected values.

nmaarnio:
Thanks for the review Nikolas, important notes. I agree what Nikolas proposed to be the focus of refactoring efforts: to generalize all wrangling specific to data formats elsewhere, focus on the core functions and add tests for them. If we remove the table/raster separation from these functions, we would be left with just two functions per transformation function. This would allow us to remove the "core"-suffix from the inner function (since there is no more 3 layers per functionality) and follow the standard naming conventions of EIS Toolkit. If we wish to keep "core" in the names of the refactored version, it raises a question should we add "core" to the names of other inner functions of EIS Toolkit too.

I will draft some general utility functions that can be used to convert raster and vector inputs to other formats and reuse parts of what you have already done Michael.

nialov:
> The _binarize_core() and the other transformation functions are already in that shape, only accepting a single array/band for transformation.

Yeah, this was a great design choice! 

> When I´m not completely wrong, the only **alternative** to avoid this and keep memory usage at lowest, would be reading every band for its own, but then also have only single band outputs (i.e., 5 files for a raster with 5 bands). Otherwise (when enabling multi-band outputs), stacking of slices will probably require similar or higher memory resources ❓

`rasterio` can write a single band at a time into raster in `DatasetWrite.write`. So you can read the single band in a multiband raster, process it, and write it back to the same band in the output raster. This way, only a single band is in memory at the same time. As far as I understand things, unless the process needs the bands available at the same time, there should be no reason to read all the bands into memory.

Memory issues should not be be waived as something related to just lacking computing hardware. Rasters can come in a wide variety of formats and file sizes. For most use cases, regardless of their size, `rasterio` can read, process and write them if used correctly. E.g. windowed reads, where only a chunk of the raster is read into a memory at a time, allows processing rasters that exceed the memory size of the computer in use. This is how e.g. `QGIS` can show rasters of terabyte size on any computer. Not saying this is a priority currently (or maybe ever in `eis_toolkit`) but I would say that most memory issues can be solved by the programmers rather than waiving the issues as users lacking hardware. So I like to point these things out as I have run into the memory issues in the past.

nialov:
> Okay, thank you 👍 , I was not aware of this 👀 feature. However, the `eis_toolkit` processing is based on a `rasterio` object as input and always returns a `np.ndarray`. This is the behavior of most other functions in the `eis_toolkit`, too:
> 
>     1. read/convert the `rasterio.io.DatasetReader`
> 
>     2. do some magic 🦄
> 
>     3. return a `np.ndarray`
> 
> 
> In my understanding, saving the output to a desired format is another seperate step and general functionality, which initially was not part of the transformation module. Keeping the input/output requirements in our function table(s), which are either `np.ndarray`s or `pd.dataframe`s, I`m not sure to solve this without at least another outer function doing the saving job. Beside, the code would be **not consistent** with the rest of the `eis_toolkit`modules (returning`np.ndarray`s instead of directly saving the results) if we change it this way.

I agree completely that the saving is a separate step, sorry if my previous comment was confusing. Just tried to elaborate the "usual" `rasterio` workflow. As far as I see all the transformation functions (and in general most `eis_toolkit` modules/functions) just return a `np.ndarray` rather than writing any files. Writing should happen after calling the functions closer to the command-line interface implementation, as you mentioned.

I think the input should be as simple as possible. For example the `DatasetReader.profile` contains most relevant metadata, including the affine transformation, so even the `DatasetReader` might not be necessary for most functions. The profile along with the read array data as inputs should cover most use cases from what I have seen so far. But this is not critical and can be refactored later if there is time.

> Another point, which came to my mind: When someone builds a model, all relevant data must be loaded into memory anyway ❓ That given, memory load will be even higher? Even if I'm not 100% sure how scikit-learn and tensorflow handle these under the hood, I think the data must be loaded into memory before doing training/inferencing.

I could not say. I suppose these are problems that we will solve when they come up. But the user always has the option to crop or resample their data to succeed in modelling. Same applies to any other functions as well of course but I believe the memory issues should still be kept in mind.

nmaarnio:
Regarding the **writing**, let's do the saving in a separate function as you both seemed to agree. I think it helps to think about the three ways the toolkit can be used:
- Using this QGIS plugin (that will use the CLI of EIS Toolkit)
- EIS Toolkit CLI directly
- Using EIS Toolkit as a Python library in scripts and software

When EIS Toolkit is used as a library, we can trust the users can call a few functions in a meaningful order as long as they are implemented. What then remains is the CLI, where we can handle tasks related to getting the data in the desired format from an input file and save the processed data back into a file. To put it short, I agree with @nialov  that writing (and reading, and related operations) should happen close to the CLI, and that it is possible to do for example band-by-band processing using the CLI without concerning the data function with these parameters.

Regarding the **memory concerns**, I think it is quite a complicated issue, and I admit I do not know a lot about this topic. But when a function is called from the plugin/CLI, EIS Toolkit is opened, it does the one ordered process (or short workflow) and closes afterwards. This means that during a modelling session there might be quite a lot of both reading data in memory and dropping data from memory. But the test result you mentioned @msmiyels seems reasonable to me. As already said, let's keep this in mind as we continue to develop the toolkit.

eis_toolkit/transformations/binarize.py
```diff
@@ -0,0 +1,101 @@
...
+    out_array = np.where(in_array <= threshold, 0, 1)
+
+    return out_array
+
+
+@beartype
+def binarize(  # type: ignore[no-any-unimported]
+    raster: rasterio.io.DatasetReader,
+    bands: Optional[Sequence[int]] = None,
+    thresholds: Sequence[Number] = [Number],
```
nmaarnio, line 29: Maybe leave the default parameter out here too? Anyway, I believe `[Number]` is not a valid default parameter

nmaarnio (commented):
Hi, I am still in the process of pondering of some things related to the utilities in the PR, but I'll give some comments already still:
- You should enable and run `pre-commit`, now there are multiple things the linter complains about
- I noticed some parameters like `percentiles` and `limits` have default values, but these values will raise an exception if I read correctly. I would simply leave defaults values out in these cases and force the user to give some meaningful parameter values
- This is minor thing and nitpicky, but it could look a bit better if both `check_raster_bands` and `check_parameter_length` use the same evaluation form, either `is False` or `not`
- Using the `|` operator in type hints might not work according to our previous research, so we should stick to `Union` I think
- The `Number | None` in several places could be changed to `Optional[Number`

Overall I think these functions are understandable and correct! I am still a bit unsure still whether it is good idea to handle nodata inside these functions and allow to process multiple bands at one time. Nodata handling and band-specific behavior needs to be defined for many other tools too. However, let's not polish these functions forever but try to merge them soon. I would still like you to give your review @nialov when you have time.

eis_toolkit/transformations/binarize.py
nmaarnio: Hi, I don't see a notebook included in the PR right now, or do you mean you tested it just locally? Anyway, when I modified the test files so that I removed `limits=limits` or `percentiles=percentiles` from the function calls (making the functions use the default values), the tests failed raising exceptions `InvalidParameterValueException: Limit values all None: (None, None).` or `InvalidParameterValueException: Percentile values all None: (None, None)`.

I think the separation makes sense if it is often the case that the user wants to give one limit/percentile but leave the other as None. Then I would go for your approach 1. Otherwise if we want to define these inputs as a tuple, I would just remove the default values and force the user to always define inputs, for example `limits = (5, None)`.

We can keep possibility to process multiple bands since it has been implemented already. Better to change it only when we really see a need for it and not discard work already done :+1:.

nialov:
I will go through the pull request this week or next.

nialov:
I will focus on the core functionality (functions prefixed with `_`) and the provided tests.

# Test modularization

-  Create a folder `tests/transformations` and move all tests there
-  In `tests/transformations/`, create a `__init__.py` file. Use this file to modularize some of the testing. E.g.

I see this pattern repeated across tests:

~~~python
# Output shapes and types
assert isinstance(out_array, np.ndarray)
assert isinstance(out_meta, dict)
assert isinstance(out_settings, dict)

# Output array (nodata in place)
test_array = raster.read(list(range(1, out_meta["count"] + 1)))
np.testing.assert_array_equal(
    np.ma.masked_values(out_array, value=nodata, shrink=False).mask,
    np.ma.masked_values(test_array, value=nodata, shrink=False).mask,
)

~~~

Move it into its own function in `tests/transformations/__init__.py` like so:

~~~python

def check_transformation_outputs(out_array, out_meta, out_settings, raster, nodata):
    # Output shapes and types
    assert isinstance(out_array, np.ndarray)
    assert isinstance(out_meta, dict)
    assert isinstance(out_settings, dict)

    # Output array (nodata in place)
    test_array = raster.read(list(range(1, out_meta["count"] + 1)))
    np.testing.assert_array_equal(
        np.ma.masked_values(out_array, value=nodata, shrink=False).mask,
        np.ma.masked_values(test_array, value=nodata, shrink=False).mask,
    )

~~~

Then you can import and use it inplace of the repeated code like so:

~~~python
from tests.transformations import check_transformation_outputs

def test_binarize():
    """Test that transformation works as intended."""
    bands = None
    thresholds = [2]
    nodata = 3.748
    nodata = cast_scalar_to_int(rasterio.open(raster_path).nodata if nodata is None else nodata)

    with rasterio.open(raster_path) as raster:
        out_array, out_meta, out_settings = binarize.binarize(
            raster=raster, bands=bands, thresholds=thresholds, nodata=nodata
        )
        check_transformation_outputs(out_array=out_array, out_meta=out_meta, ...)

~~~

This reduces duplicate code in tests. Please check if other parts can be modularized like so! Helps a lot in reading the tests as the general parts are more easily seen.

# Test rationale

In e.g. `tests/transform_binarize_test`, do I misunderstand the code or do you repeat
the calculation done by `binarize.binarize` in the test function?

~~~python
for i in range(0, out_meta["count"]):
    test_mask = np.ma.masked_values(test_array[i], value=nodata, shrink=False).mask
    test_array[i] = np.where(test_array[i] <= thresholds[i], 0, 1)
    test_array[i] = np.where(test_mask, nodata, test_array[i])
~~~

This is illogical as how can you then be sure the test code is correct? You should instead e.g. run `binarize._binarize` on a simple array, check the result is sensible/correct and then save the result e.g. in code in a test python file (See e.g. <https://github.com/GispoCoding/eis_toolkit/blob/5ccdee4543547783d11fbb0e0d68d4d978558f80/tests/vector_processing/test_vector_density.py#L18-L32> where the expected array is manually inputted into the file on lines 18-32). Then in the test check that the result is the same as the expected result (you do not have to use `pytest.mark.parametrize` as done in my example). Personally I do not see the value in repeating the analysis code in the test itself whereas checking output types and e.g. sizes of output arrays (e.g. should the dimensions be the same as in the input array?) are sensible tests.

# Test targets

I would also like for the tests to have simple examples of all the core functions. E.g. a simple test like

~~~python

def test_min_max_scaling_core():
    input_array = np.array([1, 2, 3, 5, 11])
    input_range = (0, 1)
    expected_result = np.array([0.0 , 0.1, 0.2, 0.4, 1.0])
    result = _min_max_scaling(input_array, input_range)
    assert isinstance(result, np.ndarray)
    assert len(result) == len(input_array)
    np.testing.assert_array_equal(input_array, expected_result)
~~~

would very clearly show what the function does and the test is easily editable and understandable in the future. If this test fails, you can immediately be sure that you have messed the core functionality during some refactoring.

I have do not think that there are issues in the core functions currently but these sort of tests are meant to allow safely editing them in the future! And to make reviewing easier as the functionality is plainly available from the test.

nialov:
> Hi and thank you for the comments @nialov 😃
> 
> I've changed some parts in the testing files:
> 
>     * **removed** the code for manual calculation and comparison of the main functions
> 
>     * **added** the `__init__.py` with your suggested functionality
> 
>     * **added** `test_core()`functions with an simple example
> 
> 
> However, I would not like to compare the output of the main functions with pre-calculated arrays because every point should be considered within the other tests, like `shapes`, `dtypes`, `exceptions`, nan-replacement and -positions as well as provided samples in the `_test_core()` functions. There could be plenty of stuff and variants tested, too, but I do not really see a strong benefit from more extensive testing then implemented at the moment.
> 
> Let me know if this is okay so far or if it's really needed to add further pre-calculated results and comparisons, which are basically already included in the `_test_core() ` examples.

This is fine! The added core function tests work well as the "regression" tests I was talking about. No need for further additions.

> 
> Cheers, Michael

Core and test functionality now look good to me! @nmaarnio

nmaarnio (approved):
Looks good to me, let's get this merged! I can resolve the conflicts and let's merge after that.

nialov (approved):

PR #109, opened 2023-03-15
76 add random forest functionality

BerndTorchala:
random forest functions finished and tested


nmaarnio (changes requested):
There are several issues with this pull request at the moment. To avoid putting many individual comments to the files and within code, I will try to break them down here in categories.

**Style**
Overall, it seems the style guide is not read and linter/pre-commit has not been used. There are many places where there are for example unncessary comments, commented old code, whitespaces and inadequate type hints. Please read the [contributing guide](https://github.com/GispoCoding/eis_toolkit/blob/master/CONTRIBUTING.md) carefully to fix many of the issues in this PR. Besides fixing the points mentioned in this paragraph:
- Use clear variable names.
- Use snake_case, not CamelCase in naming
- Don't use Any-type at all
- Don't use general try except
- Don't use asterisk (*) importng

**Files**
- Avoid putting in a lot of test data. At the moment there is way too much test data coming with this PR. Be careful with this, because the data clutters the repository already before merging the PR.
- Now there were modifications in files that should not relate to developing random forest functions. Please remove the modifications from this PR.
- Overall, there are too many Python files and in wrong folders in this PR. Communicating this to all developers has not been as clear as it should have so far, but we wish to limit the number of new files, even if we aim for high modularity. Ultimately, I would like this PR to have only a few files, excluding testing.
- About utility functions: If there is a clear need for a general utility function, a separate GitHub issue should be raised about this need. NoData handling is one good example. We don't want to redo these functions many times. Now me, Nikolas and Michael will implement a set of common utility functions, and if there is something more specific to random forests, it should all be in a file called "random_forest_utils.py" or something like that.

**Other**
- Now there is a lot of emphasis on input validation. I will create a decorator that will check all input arguments are of correct type, so you should remove all of these checks from your code.
- I apologize that reviewing this took so long. To ease this task, and more importantly to avoid changing a lot of the code in the future, I would ask for smaller PRs and a **minimal** implementation at first. We can always iterate the implementation and add new parameters, checks etc. as we see fit. When developing a feature, you can always tag me or someone else in the related issue to ask for comments, or raise a new issue if needed.
- If you want my honest recommendation how to continue working on this PR, I would start almost from scratch and redo the whole thing, with minimal implementation first. Just do what is really central to random forests with clean code, and let's implement all adjacent tasks separately.

nmaarnio:
Hi! When me and Nikolas reviewed the [transformation functions PR](https://github.com/GispoCoding/eis_toolkit/pull/98), we decided that we aim to separate all general utility functions, data format specific functionalities and to strictly focus on the feature itself (and tests for it). I suggest we do the same here.

So, you too could focus refactoring efforts strictly to the core random forest functionality. This would mean a dramatic decrease of code in this PR. As with Michael's transformation functions, I will grab from your code some utility functionalities and implement them in a separate PR, so you can safely remove them from here and assume adequate functions will be implemented elsewhere.

nmaarnio:
Closing this issue as basic sklearn random forest is now part of master branch (PR #210).

PR #110, opened 2023-03-21
82 add unify rasters

nmaarnio:
Added unify rasters functionality.

The function takes a base raster and a list of rasters-to-be-unified. The unified rasters will have the same CRS, alignment, pixel size and optionally extent as the base raster.

Notably, this implementation does not utilize the previously implemented reproject, resample, snap and clip functions. A direct call to rasterio.warp.reproject allowed to avoid multiple raster writing and reading and a more accurate result than was achieved with the 4 mentioned raster processing functions of EIS Toolkit.

Some modifications were made to for example Resampling and Snapping while developing this functionality. Initially, the idea was to utilize these functions which is why they were modified.


Mtk112 (approved):
Everything looks good. Merging.

Issue #102, opened 2023-03-08
Add rasterize to convert vector data to raster


nialov:
Rasterization is the process of "burning" the values to a raster where a vector geometry intersects raster cells. The burned value can either be a binary value or a value from the vector geometry. Geometries can be points, lines or polygons.

Issue #120, opened 2023-04-12
Add utility functions


nmaarnio:
A common set of utility functions are needed to perform some common tasks and to have higher level of code reuse and modularity. It has not been decided yet what exact functions will be included, but at least the following needs are addressed:
- File I/O
- NoData handling
- Common operations on dataframes and rasters
- Checks (converting checks as a submodule of utilities will be considered)

nialov:
While implementing the `rasterize` functionality, I noticed that it allows the
input of a base raster to use as the "canvas" in which to burn the rasterized
vectors in. However, only the profile of the raster is needed, not the full
`DatasetReader`. I have noticed that the full `DatasetReader` is often passed
around in `eis_toolkit` even though only the profile of the raster would be
sufficient. I propose an utility function to get the profile of a raster:

~~~python
from pathlib import Path

import rasterio
from rasterio import profiles


def raster_profile(raster_path: Path)->profiles.Profile:
    with rasterio.open(raster_path) as raster:
        profile = raster.profile
    return profile
~~~

It is simple, but it is useful to avoid having to use a context manager or
passing around the full `DatasetReader`.
    
Example `profiles.Profile` contents:

~~~
{'count': 1,
 'crs': CRS.from_epsg(3067),
 'driver': 'GTiff',
 'dtype': 'float64',
 'height': 8,
 'interleave': 'band',
 'nodata': -999.999,
 'tiled': False,
 'transform': Affine(2.0, 0.0, 384760.5,
       0.0, -2.0, 6671362.5),
 'width': 10}
~~~

nialov:
> Yes, I think we could have this as one of the utility functions. Would a similar function be useful for raster.meta, or is there any reason even to use meta instead of profile (as I understand, profile has the same information and some additional information)?

I believe profile has the most comprehensive metadata. So it should contain all `meta` attributes as well.

PR #122, opened 2023-04-13
Add rasterize

nialov:
Added rasterization of vector geometries as discussed in #102. Contains
capabilities for all geometry types and allows adding a buffer around the
geometries to increase the subsequent raster cell area that is burned by the
geometry.

Added tests for the basic geometry types, using a base raster (profile) and
buffering input geometries.

Ready for review as of 14.4.2023.

### DIFF
--- a/eis_toolkit/vector_processing/rasterize_vector.py
+++ b/eis_toolkit/vector_processing/rasterize_vector.py
@@ -43,6 +43,7 @@ def rasterize_vector(
         raise ValueError(f"Expected a positive buffer_value ({dict(buffer_value=buffer_value)})")
 
     if buffer_value is not None:
+        geodataframe = geodataframe.copy()
         geodataframe["geometry"] = geodataframe["geometry"].apply(lambda geom: geom.buffer(buffer_value))
 
     return _rasterize_vector(
--- a/eis_toolkit/vector_processing/rasterize_vector.py
+++ b/eis_toolkit/vector_processing/rasterize_vector.py
@@ -1,12 +1,13 @@
-from typing import Optional, Tuple, Union
-
 import geopandas as gpd
 import numpy as np
+from beartype import beartype
+from beartype.typing import Optional, Tuple, Union
 from rasterio import features, profiles, transform
 
 from eis_toolkit import exceptions
 
 
+@beartype
 def rasterize_vector(
     geodataframe: gpd.GeoDataFrame,
     resolution: float,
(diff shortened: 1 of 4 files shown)


eis_toolkit/vector_processing/rasterize_vector.py
```diff
@@ -0,0 +1,111 @@
...
+from eis_toolkit import exceptions
+
+
+def rasterize_vector(
+    geodataframe: gpd.GeoDataFrame,
+    resolution: float,
+    value_column: Optional[str] = None,
+    default_value: float = 1.0,
+    fill_value: float = 0.0,
+    base_raster_profile: Optional[profiles.Profile] = None,
```
nialov, line 16: The main entrypoint takes in a `rasterio.profiles.Profile` rather than a `rasterio.io.DatasetReader` which is more often the case in `eis_toolkit`. I mentioned this in #120. I believe it is appropriate in just this function as the array data from the raster is not needed but there might be some value in always passing the reader instead to avoid people having to figure out what is the `profiles.Profile`. Small details of course...

tests/vector_processing/test_rasterize_vector.py
```diff
@@ -0,0 +1,258 @@
...
+    base_raster_profile: Optional[Union[profiles.Profile, dict]] = None
+    buffer_value: Optional[float] = None
+    raises: ContextManager = nullcontext()
+
+
+@pytest.mark.parametrize(
+    "geodataframe,resolution,value_column,default_value,fill_value,base_raster_profile,buffer_value,raises",
+    [
+        pytest.param(
+            *RasterizeVectorTestArgs(geodataframe=SAMPLE_LINE_GEODATAFRAME, resolution=0.05),
```
nialov, line 84: I took some liberties in handling test parametrization. I used a `NamedTuple` to represent the test arguments along with their defaults. This makes refactoring later easier as the order of the test arguments along with their defaults can be only changed in `RasterizeVectorTestArgs` rather than having to do it separately for each tuple of test arguments. Hopefully not too complex...

nialov:
What I realized while reviewing pull request #98 that a general rasterization function for fixed grid point data does not exist yet. I think this pull request could be merged with the general rasterization functionality for any geometries and a separate pull request for specifically for "fixed grid rasterization for point data" should be done. There the rasterization is a different process all together as the points are expected to raster cell center points rather than points as themselves.

eis_toolkit/vector_processing/rasterize_vector.py
```diff
@@ -0,0 +1,125 @@
...
+            If None, then default_value is used for all geometries.
+        default_value (float): Default value burned into raster cells based on geometries.
+        base_raster_profile (Optional[Union[rasterio.profiles.Profile, dict]]): Base raster profile
+            to be used for determining the grid on which vectors are burned in.
+        fill_value (float): VAlue used outside the burned/rasterized geometry cells.
+        buffer_value (Optional[float]): For adding a buffer around passed
+            geometries before rasterization.
+
+    Returns:
+        out_raster (tuple(numpy.ndarray, dict)): Rasterized vector data and metadata.
```
nmaarnio, line 34: You could remove the type annotations and output value name from here. There seems to be also a typo on line 29: "VAlue" instead of "Value".

eis_toolkit/vector_processing/rasterize_vector.py
```diff
@@ -0,0 +1,125 @@
...
+
+    out_raster_array = features.rasterize(
+        shapes=geometry_value_pairs,
+        # fill and default_value can be floats even though typing claims otherwise
+        fill=fill_value,
+        default_value=default_value,
+        transform=out_transform,
+        out_shape=(round(height), round(width)),
+    )
+    return out_raster_array, dict(transform=out_transform, height=height, width=width)
```
nmaarnio, line 125: I wonder should we return more metadata here? We could add at least "count", "driver", "crs" and "dtype".

nialov: I believe the question is if the output raster file needs to be composeable from the output dictionary/profile. If we require the individual functions to collect the needed keys and values into a working `rasterio.profiles.Profile` or `dict`, we should have an utility function to create it and check the output rather than trusting that each module properly returns the right keys and values. 

I only included `transform` and the dimensions as they are the only part that is only known within the function. `crs` can be extracted from the `geodataframe`, `count` is always 1, `driver` depends on what file we want to create and `dtype` can be extracted from the `out_raster_array`.

nmaarnio: Ok, perhaps this is best to leave as it is now. We can change it later if needed, or create a utility function as you mentioned.

nmaarnio (approved):

PR #126, opened 2023-04-21
Add distance computations (Closes #103)

### LINKED ISSUE #103: Add distance computations
Distance computations consists of calculating the distance to the nearest vector geometry for each raster cell. This distance is added as the value of the raster cell. The "empty" raster in which the values are calculated should be provided by the user or created from user inputs (e.g. bounds and cell size).

nialov:
The function uses the base raster profile to create vector points in the center
of each raster cell. It then calculates the minimum distance to the provided
geometries for each point and consequently adds it as the output raster cell
value.

### DIFF
--- a/eis_toolkit/spatial_analyses/distance_computation.py
+++ b/eis_toolkit/spatial_analyses/distance_computation.py
@@ -1,0 +1,82 @@
+from functools import partial
+
+import geopandas as gpd
+import numpy as np
+from rasterio import profiles, transform
+from shapely.geometry import Point
+
+from eis_toolkit import exceptions
+
+
+def _smallest_distance_to_point(point: Point, geometries: gpd.GeoDataFrame) -> float:
+    """Calculate smalled distance from geometries to point."""
+    # This could be sped up by using a spatial index to first choose only
+    # geometries that are close to the point
+    distances_to_point = geometries.distance(point)
+    smallest_distance = distances_to_point.min()
+    return smallest_distance
+
+
+def distance_computation(raster_profile: profiles.Profile, geometries: gpd.GeoDataFrame) -> np.ndarray:
+    """Calculate distance from raster cell to nearest geometry.
+
+    Args:
+        raster_profile: The raster profile of the raster in which the distances
+            to the nearest geometry are determined.
+        geometries: The geometries to determine distance to.
+
+    Returns:
+        A 2D numpy array with the distances computed.
+
+    """
+    if raster_profile.get("crs") != geometries.crs:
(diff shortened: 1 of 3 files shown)


eis_toolkit/spatial_analyses/distance_computation.py
```diff
@@ -0,0 +1,81 @@
...
+
+from eis_toolkit import exceptions
+
+
+def _smallest_distance_to_point(point: Point, geometries: gpd.GeoDataFrame) -> float:
+    """Calculate smalled distance from geometries to point."""
+    # This could be sped up by using a spatial index to first choose only
+    # geometries that are close to the point
+    distances_to_point = geometries.distance(point)
+    smallest_distance = distances_to_point.min()
```
Mtk112, line 14: To fix no-any-return linter error:
```suggestion
    shortest_distance: float = distances_to_point.min()
```

Mtk112 (changes requested):
The code looks good to me, however few nitpicky notes and general suggestions relating to linter.

nialov:
Thanks for the review @Mtk112 ! I am interested in if using `disallow_any_unimported = True` is worth setting in `mypy.ini` if we then have to annotate most functions with the `#type ignore...` comments to disable it? I believe most libraries we use are quite lacking in type annotations (e.g. `numpy`, `geopandas`, `pandas`, etc.) so reinforcing the rule is quite difficult.

eis_toolkit/spatial_analyses/distance_computation.py
```diff
@@ -0,0 +1,81 @@
+import geopandas as gpd
+import numpy as np
+from rasterio import profiles, transform
+from shapely.geometry import Point
+
+from eis_toolkit import exceptions
+
+
+def _smallest_distance_to_point(point: Point, geometries: gpd.GeoDataFrame) -> float:
+    """Calculate smalled distance from geometries to point."""
```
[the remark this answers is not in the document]
nialov: Renamed and added the type comment.

eis_toolkit/spatial_analyses/distance_computation.py
nialov: Fixed!

nmaarnio:
> I am interested in if using `disallow_any_unimported = True` is worth setting in `mypy.ini` if we then have to annotate most functions with the `#type ignore...` comments to disable it?

I just thought about this too yesterday when making a refactoring/style change PR, and I agree we could remove it. I will make another commit to my PR and remove it there, if nobody objects.

nialov:
Removed the `# type: ignore` annotations!

PR #127, opened 2023-05-02
Add cba

AlVella:
## Addition of Cell-Based Association

This PR contains tools necessary to perform CBA:
- CBA main code.
- Notebook with a short introduction on how to process data with the CBA.

CBA allows to process geological data using a regular square-cell matrix. This grid is joined spatially with geological data (e.g. lithologies, linear structures, point occurences). Each geological object is translated into a binary variable in a unique matrix. For each cell, the absence/presence of each object is translated by a "0" or a "1" respectively.

Unique environments can be drawn from those binary spectrum, some of them associated with known mineralization. Those environments are then used to produce favorability maps using various predictive methods (e.g. "Ranking", AHC, or Random Forests).

CBA **does not** take into account the surface of lithologies or the number of occurrences, only their presence or absence. Even if a specific object is barely present in a cell, it will be translated by a "1".

## Summary

The CBA is implemented as a class. All necessary functions are implemented inside this class.

Instantiating the CBA requires .shp inputs. It is **mandatory** to use lithological dataset to initiate the CBA. Structural data or known deposits/occurrences must be added after.

Gridding numerotation goes rows by rows, West to East and North to South. The first cell is on the NW corner and the last cell on  the SE corner.

Results are stored in the attributes of the class and can be exported as a .shp file or as .csv file.

Results can be re-instantiated with in-built functions if necessary (e.g. adding new data to the matrix), from both .csv or .shp files.

All inputs data **must have** the same CRS as the initial dataset used to instantiate the CBA. The GeoDataFrame produced by the CBA will use the CRS of the dataset used to instanciate it. Adding new data to the CBA with different coordinate system can cause discrepancies in the results or even fatal error.


nialov (changes requested):
Took a look at this pr. Some suggestions below. Let me know if something is
unclear!

# Documentation is excellent

The notebook in particular is great in explaining the functionality very
clearly. However, I noticed that running the notebook modifies
`tests/data/remote/Test_CBA_matrix.dbf`. I would recommend not writing
any data that is included in the repository in the notebook. Writing
should be done to some temporary directory. E.g.

``` python
from tempfile import TemporaryDirectory
from pathlib import Path

with TemporaryDirectory() as tmp_dir:
   output_file = Path(tmp_dir / "output.shp")
   cba_grid.to_shapefile(str(output_file))
```

# There is significant overlap with the functionality of `rasterize`

As I understand it cell-based association rasterizes each category of
e.g. lithology or structural feature type separately with binary values
(0 or 1). I believe you could loop over each category and use
`rasterize` to create a matrix with ones and zeroes. After looping over
each category, the matrices could be joined into a multi-band raster
where each band represents each category. The result would therefore be
similar to the vector results presented in this implementation as
instead of columns, the bands would represent the different categories.
My understanding might be a simplification, this is just how I have
understood the basic case of *CBA*.

Another advantage would be that the results are already in raster format
which is a requirement for the "modelling" phase. From what I
understand, the results of *CBA* is always a fixed grid where each cell
is the same size? This kind of data would be better represented as a
raster.

However, this implementation also enables using circular disks around
the grid cells for the rasterization which is not provided by
`rasterize`. I.e. instead of using the rectangular grid cells to
determine intersections, a circle is created from the centroid of the
rectangular grid cell that is used to find intersecting geometries. This
functionality is new for `eis_toolkit`.

**In summary:** refactoring to using `rasterize` (existing `eis_toolkit`
functionality) could be considered but parts of *CBA* are novel (disks)
and can not be replaced with functionality from `rasterize`. I suggest
going with the current implementation for now as it seems well coded and
is working.

# Results are not provided in raster format

The results are provided in a `geopandas.GeoDataFrame` which has Polygons in
a fixed grid and columns where the values of cell-based association are
provided. This should be converted to either a multi-band raster where each
band is a matrix with the values from one column or to multiple single-band
rasters. Or a collection of `numpy` arrays.

# Use snake_case for variable names

-   <https://en.wikipedia.org/wiki/Snake_case>

This is a code style thing. All variables should start with lower-case
letters. I see uses of e.g. `Name`, `CentX` and `Disc`. These should be
renamed to `name`, `cent_x` and `disc`, respectively.

# Test data as shapefiles

Personally I prefer using `GeoJSON` for test data as it is text rather
than binary data and therefore a bit more suitable for `git`, especially
is formatted with newlines. This is just a small detail to consider if
wanted.

eis_toolkit/spatial_analyses/cba.py
```diff
@@ -0,0 +1,511 @@
...
+    grid = get_grid(geodata, cell_size, cell_size)
+    grid.crs = geodata.crs
+
+    # Filling the grid with overlapped geometries
+    indicators, join_grid = prepare_grid(geodataframe, grid, column, subset_of_target_attribute_values)
+    tmp = join_grid[["geom"]].copy(deep=True)
+    tmp.rename({"geom": "geometry"}, axis=1, inplace=True)
+
+    # Saving results in attributes
+    cba = gpd.GeoDataFrame(pd.concat([tmp.groupby(tmp.index).first(), indicators], axis=1))
```
nialov, line 70: Right now the Python module is `cba.py` and within the code `cba` is used as the variable name for the `GeoDataFrame` holding the data. This can cause mistakes when the module is used. I would rename either the module (e.g. `cba.py` -> `cell_based_association.py`, I prefer this option!) or the `cba` variable in code (e.g. `cba` -> `cba_data`).

eis_toolkit/spatial_analyses/cba.py
```diff
@@ -0,0 +1,511 @@
...
+
+def prepare_disc(  # type: ignore[no-any-unimported]
+    geodata: gpd.GeoDataFrame,
+    disc: gpd.GeoDataFrame,
+    target_attribut: str,
+    subset_of_target_attribut_values: list,
+) -> Tuple[list, gpd.GeoDataFrame]:
+    """Intermediate utility.
+
+    Compute spatial joint and generate the table of presence of the different
```
nialov, line 305:
```suggestion
    Compute spatial join and generate the table of presence of the different
```

eis_toolkit/spatial_analyses/cba.py
```diff
@@ -0,0 +1,511 @@
...
+        geodata: the geodata object corresponding to the vector file.
+        disc: existing disk 'paving'.
+        target_attribut: target attribute to extract from the vector file.
+        subset_of_target_attribut_values: subset of the relevant values
+            of the target attribute
+
+    Returns: pair of variables
+        -> boolean indicating if the target attribute must be dummified
+        -> grid resulting from the intersection between the mesh and the
+            vector file
```
nialov, line 318:
```suggestion
    Returns: A boolean indicating if the target attribute must be dummified
        and a intersection grid between the mesh and the vector file
```

eis_toolkit/spatial_analyses/cba.py
```diff
@@ -0,0 +1,511 @@
...
+                    (x_right_origin, y_bottom),
+                    (x_left_origin, y_bottom),
+                ]
+            )
+            data += [[i + j * cols, i, j, polygon]]
+            y_top = y_top - cell_height
+            y_bottom = y_bottom - cell_height
+        x_left_origin = x_left_origin + cell_width
+        x_right_origin = x_right_origin + cell_width
+    result = pd.DataFrame(data, columns=["cell_id", "x", "y", "geom"])
```
nialov, line 381: Unless there is a specific reason, I would not allow the `"geom"` column name in any outputs from any of the functions. As I understand, there is always just one geometry column for each `gpd.GeoDataFrame` that is outputted from the functions. I would therefore just use the standard `"geometry"` column name. If multiple geometry columns exist internally at some point, it is fine to differentiate between them, but the `"geom"` column should not be exposed outside the functions as it is confusing. I also encountered errors in the `cba_introduction.ipynb` where a `pd.DataFrame` had a `"geom"` column but was not a `GeoDataFrame`.

eis_toolkit/spatial_analyses/cba.py
```diff
@@ -0,0 +1,511 @@
...
+    xmin, ymin, xmax, ymax = working_map.total_bounds
+    rows = int(np.ceil((ymax - ymin) / cell_height))
+    cols = int(np.ceil((xmax - xmin) / cell_width))
+    x_left_origin = xmin
+    x_right_origin = xmin + cell_width
+    y_top_origin = ymax
+    y_bottom_origin = ymax - cell_height
+    print("Columns: %s" % cols)
+    print("Rows: %s" % rows)
+    print("Cells: %s" % (cols * rows))
```
nialov, line 362: You can use `logging` to report runtime info. `print` should be avoided as it cannot be controlled outside the function.

```suggestion
    import logging # at top of the module
   
    logging.info("Columns: %s" % cols)
    logging.info("Rows: %s" % rows)
    logging.info("Cells: %s" % (cols * rows))
```

Or alternatively just remove the `print` code.

eis_toolkit/spatial_analyses/cba.py
```diff
@@ -0,0 +1,511 @@
...
+    transform = rasterio.transform.from_bounds(min_x, min_y, max_x, max_y, width=width, height=height)
+
+    with rasterio.open(
+        output_path + ".tif",
+        mode="w",
+        driver="GTiff",
+        height=height,
+        width=width,
+        count=count,
+        dtype=cba[col].dtype,
```
nialov, line 502: Currently this just picks the dtype from the last column in the `GeoDataFrame`. During execution of the `cba_introduction.ipynb` notebook I noticed that the last column was named `geom` and had dtype of `object`. This causes the `rasterio.open` to fail as `object` is not a valid `dtype`. Can the `dtype` be explicitly set to `int64` as the data to be written is always integers (0 or 1)? Also, you should consider filtering the columns using a dtype (or dtypes) rather than column names (`drop("geometry")`) as e.g. the `geom` column got passed through.

```suggestion
        dtype="int64",
```

notebooks/cba_introduction.ipynb
```diff
@@ -0,0 +1,1091 @@
...
+      "text/plain": [
+       "<Figure size 640x480 with 1 Axes>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "shp_matrix = cba.add_layer(\n",
```
nialov, line 1007: I got an error when running the notebook here:

~~~
AttributeError: 'DataFrame' object has no attribute 'dissolve'
~~~

This is due to the `grid` object being a `DataFrame`. It has a `geom` column but has not been for some reason not been converted to `GeoDataFrame`.

eis_toolkit/spatial_analyses/cba.py
```diff
@@ -0,0 +1,511 @@
...
+    X, Y = np.meshgrid(x, y)
+    x_resolution = X[0][1] - X[0][0]
+    y_resolution = Y[1][0] - Y[0][0]
+    min_x, min_y, max_x, max_y = cba.total_bounds
+    width = (max_x - min_x) / x_resolution
+    height = (max_y - min_y) / y_resolution
+
+    values = []
+    col_name = []
+    for col in cba.columns.drop("geometry"):
```
nialov, line 487: Due to the use of `geom` column name, only dropping `geometry` is no sufficient as the `cba` `GeoDataFrame` can contain the `geom` column with `object` as dtype.

```suggestion
    for col in cba.columns.drop(labels=["geometry", "geom"]):
```

Alternatively, you can refactor out of using the `geom` column as suggested in another comment.

tests/data/remote/Test_CBA_matrix.qmd
nialov, line 1: I do not believe this file is needed?

tests/data/remote/Test_Faults.qmd
nialov, line 1: I do not believe this file is needed?

tests/data/remote/Test_Litho.qmd
nialov, line 1: I do not believe this file is needed?

tests/raster_processing/clip_test.py
nialov, line 1: This is a file permission change, you can remove it from the pull request if you know how. Do not worry about it otherwise.

nialov (changes requested):
You should test running the `cba_introduction.ipynb` and make sure it runs successfully without references to absolute paths on your computer. Most pointed out errors are related to me trying to run it. After you get it working this pull request should be pretty much done!

eis_toolkit/spatial_analyses/cba.py
```diff
@@ -0,0 +1,511 @@
...
+    cba = cba.join(dummies).replace(np.nan, 0)
+    if column == "":
+        cba[dummies.name] = cba[dummies.name].map(int)
+
+    cba = cba.astype("int", errors="ignore")
+
+    return cba
+
+
+def prepare_grid(  # type: ignore[no-any-unimported]
```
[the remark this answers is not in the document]
AlVella: Ok, I have removed them.

notebooks/cba_introduction.ipynb
```diff
@@ -0,0 +1,1091 @@
...
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "os.environ['USE_PYGEOS'] = '0'\n",
+    "import geopandas as gpd\n",
+    "import sys\n",
+    "sys.path.append('..')\n",
+    "# import eis_toolkit.spatial_analyses.cba as cba\n",
+    "\n",
+    "sys.path.append(r\"C:\\Users\\vella\\Desktop\\EIS_Toolkit\\eis_toolkit\\eis_toolkit\\spatial_analyses\")\n",
```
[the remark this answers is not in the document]
AlVella: I forgot to do this modification. I have erased all absolute path references in the notebook and the test file, there should be only relative paths left.

eis_toolkit/spatial_analyses/cba.py
AlVella: I renamed both `cba.py` and `cba_test.py` to `cell_based_association.py` and `cell_based_association_test.py` respectively. As you raised, it should be more coherent than having duplicate names.

eis_toolkit/spatial_analyses/cba.py
AlVella: Done.

eis_toolkit/spatial_analyses/cba.py
AlVella:
```suggestion
    Returns: A boolean indicating if the target attribute must be dummified
        and an intersection grid between the mesh and the vector file.
```

eis_toolkit/spatial_analyses/cba.py
AlVella: I edited out the `"geom"` column. The only reference column for geometry is now `"geometry"` for all `GeoDataFrame`.

eis_toolkit/spatial_analyses/cba.py
AlVella: I chose the latter. Those information should be easy to retrieve from the shapefile produced or even the tiff.

eis_toolkit/spatial_analyses/cba.py
AlVella: Even with the `geom` column out of the algorithm, it would be good to avoid `geometry` type column in this function. This implementation can solve it I think.

```suggestion
    for col in cba.select_dtypes(exclude=['geometry']).columns:
```

notebooks/cba_introduction.ipynb
AlVella: Never had this error on my side. I hope the latest modification will remove it, there shouldn't be any conflict between a `geom` and a `geometry` column right now.

tests/data/remote/Test_CBA_matrix.qmd
AlVella: Seems to work fine without it. I will remove the **.qmd** files.

tests/raster_processing/clip_test.py
AlVella: Don't know how to, I'll let it be then.

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,511 @@
...
+
+    return cba
+
+
+def prepare_grid(
+    geodata: gpd.GeoDataFrame,
+    grid: gpd.GeoDataFrame,
+    target_attribut: str,
+    subset_of_target_attribut_values: list,
+) -> Union[list, gpd.GeoDataFrame]:
```
nialov, line 163: This should be `Tuple` instead of `Union`. Furthermore, a `list` is not returned from the function, rather `pandas` objects:
```suggestion
) -> Tuple[Union[pd.DataFrame, pd.Series], gpd.GeoDataFrame]:
```

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,511 @@
...
+    join_grid = gpd.sjoin(grid, geodata, how="inner", predicate="intersects")
+    if dummification:
+        tmp = pd.get_dummies(join_grid[join_grid[target].isin(identified_values)][[target]], prefix="", prefix_sep="")
+        indicators = tmp.groupby(tmp.index).max()
+    else:
+        tmp = join_grid[target]
+        indicators = tmp.groupby(tmp.index).max()
+    return indicators, join_grid
+
+
```
nialov, line 197:
```suggestion
@beartype
```

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,511 @@
...
+        raise ValueError(
+            "target_attribut != '' and subset_of_target_attribut_values == [] (numerical/string) "
+            "FORBIDDEN AT THIS TIME as not managed !"
+        )
+    else:
+        # Case of a symbolic attribute with identified subset of values of
+        # interest.
+        dummification = True
+        if attribut_values == "all":
+            attribut_values = geodata[target].unique()
```
nialov, line 253: I do not see the reason for using the `"all"` variable for indicating that all categorical values are wanted. Instead, an empty list could suffice. If the "all" option is required, the option of passing a string into `attribut_values` should be added to all function signatures.

```suggestion
    else:
        # Case of a symbolic attribute with identified subset of values of
        # interest.
        dummification = True
        if len(attribut_values) == 0:
            attribut_values = geodata[target].unique()
```

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,511 @@
...
+    assert target in geodata.columns
+    identified_values = pd.Series(attribut_values)
+    assert identified_values.isin(geodata[target]).all()
+    # FORBIDDEN AT THIS TIME as not managed !!!
+    # assert target == "" or len(attribut_values) != 0
+    assert len(attribut_values) != 0
+
+    return dummification, target, identified_values
```
nialov, line 261: Either change the signature to not need to return a list for `identified_values` or cast it as follows into a list:

```suggestion
    return dummification, target, list(identified_values)
```

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,511 @@
...
+
+
```
nialov, line 263:
```suggestion
@beartype
```

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,511 @@
...
+    return gpd.GeoDataFrame(disc, geometry="geometry", crs=geodata.crs)
+
+
+def prepare_disc(
+    geodata: gpd.GeoDataFrame,
+    disc: gpd.GeoDataFrame,
+    target_attribut: str,
+    subset_of_target_attribut_values: list,
+) -> Tuple[list, gpd.GeoDataFrame]:
```
nialov, line 305:
```suggestion
@beartype
def prepare_disc(
    geodata: gpd.GeoDataFrame,
    disc: gpd.GeoDataFrame,
    target_attribut: str,
    subset_of_target_attribut_values: list,
) -> Tuple[pd.Series, gpd.GeoDataFrame]:
```

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,511 @@
...
+    Args:
+        output_csv_file_path: name of the saved file suffixed with the
+            CRS (spatial projection system).
+
+    Returns:
+        None
+    """
+
+    crs_txt = f"epsg:{cba.crs.to_epsg()}"
+    tmp = pd.DataFrame(cba)
```
nialov, line 452: Without `copy`, the input `GeoDataFrame` is modified as a side effect of this function.

```suggestion
    tmp = pd.DataFrame(cba.copy())
```

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,511 @@
...
+    x_resolution = X[0][1] - X[0][0]
+    y_resolution = Y[1][0] - Y[0][0]
+    min_x, min_y, max_x, max_y = cba.total_bounds
+    width = (max_x - min_x) / x_resolution
+    height = (max_y - min_y) / y_resolution
+
+    values = []
+    col_name = []
+    # for col in cba.columns.drop("geometry"):
+    for col in cba.select_dtypes(exclude=["geometry"]).columns:
```
nialov, line 487: I believe this is incorrect usage, you should specify the dtypes rather than column names:

```suggestion
    for col in cba.select_dtypes(include=["int32", "int64"]).columns:
```

tests/spatial_analyses/cell_based_association_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+vector_file = gpd.GeoDataFrame.from_file(vector_path)
+points_file = gpd.GeoDataFrame.from_file(points_path)
+lines_file = gpd.GeoDataFrame.from_file(lines_path)
+
+
+def test_crs():
+    """Test that Coordinate Systems for input dataset and output grids are equivalent."""
+    grid, cba_grid = cba.init_from_vector_data(
+        cell_size=4000, geodataframe=vector_file, column="Litho", subset_of_target_attribute_values="all"
```
nialov, line 29: If the option of passing `"all"` is removed as suggested in other comment, these all should be changed accordingly:

```suggestion
        cell_size=4000, geodataframe=vector_file, column="Litho", subset_of_target_attribute_values=None
```

eis_toolkit/spatial_analyses/cba.py
nialov: I believe this is still incorrect as `'geometry'` is not a defined dtype. See my new review below.

eis_toolkit/spatial_analyses/cell_based_association.py
AlVella: Both ways works in my side. From what I understand, "geometry" can be used as a dtype with GeoDataFrame. I tried it with multiple geometry columns with various names and all were excluded.

nialov:
I cannot promise when I will be able to review the new changes, sorry! :( I might have time next week. I will do it before the end of the month in any case. But in any case it would be beneficial if @nmaarnio would like to have a quick look through the pull request in the current state of it?

nmaarnio:
I will take a look tomorrow or on Thursday!

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,573 @@
...
+
+@beartype
+def cell_based_association(
+    cell_size: int,
+    geodata: List[gpd.GeoDataFrame],
+    output_path: str,
+    column: Optional[List[str]] = [""],
+    subset_target_attribute_values: Optional[List[Union[None, list]]] = [None],
+    add_name: Optional[List[Union[str, None]]] = [None],
+    add_buffer: Optional[List[Union[Number, bool]]] = [False],
```
nmaarnio, line 24: The arguments here that have some lists as default values (such as [None] or ["]) should have booleans as default values. Optional type means that the parameter is either None or the type inside the Optional, now if user gives None things will brake down. Also, if you have these lists as default values and modify these variables (such as column) directly, unwanted and -expected things can happen if these functions are called multiple times. So, I would suggest to change the function signature to this:
```suggestion
def cell_based_association(
    cell_size: int,
    geodata: List[gpd.GeoDataFrame],
    output_path: str,
    column: Optional[List[str]] = None,
    subset_target_attribute_values: Optional[List[Union[None, list]]] = None,
    add_name: Optional[List[Union[str, None]]] = None,
    add_buffer: Optional[List[Union[Number, bool]]] = None,
```

Then, before anything else in the code, you can set the starting values like this:
`if column is None:
        column = [""]`

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,573 @@
...
+
+    Args:
+        cell_size: Size of the cells.
+        geodata: GeoDataFrame to create the CBA matrix. Additional
+            GeoDataFrame(s) can be imputed to add to the CBA matrix.
+        output_path: Name of the saved .tif file.
+        column: Name of the column of interest. If no attribute is specified,
+            then an artificial attribute is created representing the presence
+            or absence of the geometries of this file for each cell of the CBA
+            grid. If a target attribute is ndicated, it must be categorical.
```
nmaarnio, line 41: Typo here, "ndicated". Also, I think I get what the categorical attribute means here, but why could not a numerical or string type of attribute be categorical as is stated below? If the column has only for example values "1", "2", "3", or 1, 2, 3.

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,573 @@
...
+    # Allows a non-mutable default value
+    if subset_target_attribute_values is None:
+        subset_target_attribute_values = []
+
+    # Reading the vector file
+    geodata = geodataframe.copy(deep=True)
+
+    # Initialization of the grid
+    grid = _get_grid(geodata, cell_size, cell_size)
+    grid.crs = geodata.crs
```
nmaarnio, line 125: In _get_grid, the crs seems to be set to this same crs already, so this line could be removed for clarity

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,573 @@
...
+    # Spatial join of the intersection between the grid and the current map
+    join_grid = gpd.sjoin(disc, geodata, how="inner", predicate="intersects")
+
+    if dummification:
+        tmp = pd.get_dummies(join_grid[join_grid[target].isin(identified_values)][[target]], prefix="", prefix_sep="")
+        indicators = tmp.groupby(tmp.index).max()
+    else:
+        tmp = join_grid[target]
+        indicators = tmp.groupby(tmp.index).max()
+    return indicators, join_grid
```
nmaarnio, line 390: _prepare_disc and _prepare_grid seem to be the same function, so you could remove one of them.

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,573 @@
...
+        # interest.
+        dummification = True
+        if subset_target_attribute_values == []:
+            subset_target_attribute_values = geodata[column].unique()
+    assert column in geodata.columns
+    identified_values = pd.Series(subset_target_attribute_values)
+    assert identified_values.isin(geodata[column]).all()
+    # FORBIDDEN AT THIS TIME as not managed !!!
+    # assert target == "" or len(attribut_values) != 0
+    assert len(subset_target_attribute_values) != 0
```
nmaarnio, line 315: So far we have not used asserts in any other place expect tests. These could be refactored using if-statements and then raising a custom exception as in our other functions.

eis_toolkit/spatial_analyses/cell_based_association.py
```diff
@@ -0,0 +1,573 @@
...
+        add_buffer: Allow the use of a buffer around shapes before the
+            intersection with CBA cells for the added GeoDataFrame(s). Minimize
+            border effects or allow increasing positive samples (i.e. cells
+            with mineralization). The size of the buffer is computed using the
+            CRS (if projected CRS in meters: value in meters).
+
+    Returns:
+        CBA matrix is created.
+    """
+
```
nmaarnio, line 61: You could check here that cell size is valid (positive number). This would be a logical place to check also for example that all given columns exist in the geodataframe, but I noticed you test it a bit later.

tests/spatial_analyses/cell_based_association_test.py
nmaarnio, line 1: At the moment, none of the tests call the public function of CBA. I don't think you necessarily need to remove these tests, but you should include some that use the public function that users will also call. Also, check that bad data correctly raises exceptions too (invalid column name, invalid cell size at least).

nmaarnio (changes requested):

nmaarnio:
Just to let you know @AlVella , let's merge this finally to the master branch after you have done the last updates I suggested (or you can leave some out of you disagree)!

eis_toolkit/spatial_analyses/cell_based_association.py
AlVella: By default, the dummification process does not work with the integer values. I added a step to the `_add_layer()` and `_init_from_vector_data()` functions to convert targeted column to the categorical dtype, so numerical values should be working too now.

eis_toolkit/spatial_analyses/cell_based_association.py
AlVella: I added a few checks at the beginning of the `cell_based_association()` function to verify input values (no empty geodataframe, no negative cell size or buffer size ...).

tests/spatial_analyses/cell_based_association_test.py
AlVella: I updated the test file with new tests. There is now one for each exception that should be raised and another one to verify the creation of a correct .tiff grid (based on the example shown at the end of the notebook).

nmaarnio (approved):
Looks good now, thanks @AlVella ! I added rtree as a separate PR before merging this to avoid any conflicts and env problems. I will merge now!

PR #136, opened 2023-06-13
Add DBSCAN function and tests

msorvoja:


eis_toolkit/exploratory_analyses/dbscan.py
```diff
@@ -0,0 +1,62 @@
...
+    dbscan = DBSCAN(eps=max_distance, min_samples=min_samples)
+
+    dbscan.fit(coordinates)
+    data["cluster"] = dbscan.labels_
+
+    core_indices = dbscan.core_sample_indices_
+    mask = np.zeros(len(coordinates), dtype=int)
+    mask[core_indices] = 1
+    data["core"] = mask
+
```
nmaarnio, line 21: Here I would rename the "mask" to something that describes the array better. Perhaps "core_data" or something

eis_toolkit/exploratory_analyses/dbscan.py
```diff
@@ -0,0 +1,62 @@
...
+def dbscan(data: gdp.GeoDataFrame, max_distance: float = 0.5, min_samples: int = 5) -> gdp.GeoDataFrame:
+    """
+    Perform DBSCAN clustering on the input data.
+
+    Args:
+        data: GeoDataFrame containing the input data.
+        max_distance: The maximum distance between two samples for one to be considered as in the neighborhood of
+            the other. Optional parameter.
+        min_samples: The number of samples in a neighborhood for a point to be considered as a core point.
+            Optional parameter.
```
nmaarnio, line 35: These arguments are not optional, they just have a default value. So you should replace "Optional parameter" with "Defaults to X" type of description.

eis_toolkit/exploratory_analyses/dbscan.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Returns:
+        GeoDataFrame containing assigned cluster labels and value 1 if a data point is a core point
+        and 0 otherwise.
```
nmaarnio, line 39: I would make this description slightly clearer like this:
```suggestion
        GeoDataFrame containing two new columns: one with assigned cluster labels and one indicating whether a point is a core point (1) or not (0).
```

nmaarnio (commented):
Looks good, just a couple minor things about naming and docstring!

nmaarnio (approved):

PR #138, opened 2023-06-26
Add density computations

nialov:
Simply consists of small added modularization to the already done `rasterize_vector.py` file and a new simple entrypoint for the density computation (`vector_density.py`) that uses the more modularized `rasterize_vector.py` and already existing functionality.


nialov:
Seeing as "density" is the wanted characteristic, will add option to output already calculated density instead of the count which is currently returned.

nmaarnio (approved):
Looks good to me!

PR #160, opened 2023-07-03
132 add simple IDW

Mtk112:
Add simple inverse distance weighting interpolation.


tests/vector_processing/test_simple_idw.py
```diff
@@ -0,0 +1,137 @@
+from pathlib import Path
+
+import geopandas as gpd
+import numpy as np
+import pytest
+import rasterio
+from shapely.geometry import Point
+
+from eis_toolkit import exceptions
```
msorvoja, line 9: Same here as in `simple_idw.py`

msorvoja (commented):
Looks good! Left a few comments/suggestions.

eis_toolkit/vector_processing/simple_idw.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Args:
+        geodataframe: The vector dataframe to be interpolated.
+        target_column: The column name with values for each geometry.
+        resolution: The resolution i.e. cell size of the output raster.
+        extent: The extent of the output raster.
+            If None, calculate extent from the input vector data.
+        power: The value for determining the rate at which the weights decrease.
+            As power increases, the weights for distant points decrease rapidly.
+            Defaults to 2.
```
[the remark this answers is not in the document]
Mtk112: I decided to address the redundancy a bit differently.

eis_toolkit/vector_processing/simple_idw.py
```diff
@@ -0,0 +1,105 @@
...
+    extent: Union[Tuple[Number, Number, Number, Number], None],
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+
+    if geodataframe.shape[0] == 0:
+        raise EmptyDataFrameException("Expected geodataframe to contain geometries.")
+
+    if target_column not in geodataframe.columns:
+        raise InvalidParameterValueException(
+            f"Expected value_column ({target_column}) to be contained in geodataframe columns."
+        )
```
msorvoja, line 26: Maybe move exceptions into the outer function? And then add them into the docstring.

eis_toolkit/vector_processing/simple_idw.py
```diff
@@ -0,0 +1,108 @@
+from numbers import Number
+
+import geopandas as gpd
+import numpy as np
+from beartype import beartype
+from beartype.typing import Optional, Tuple
+
+from eis_toolkit import exceptions
```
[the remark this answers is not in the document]
nmaarnio: I can comment on this that it is fine to import exceptions like that, it is different from importing all contents of exceptions (not good practice): `from eis_toolkit.exceptions import *`. We have been using both `from eis_toolkit import exceptions` and `from eis_toolkit .exceptions import InvalidParameterValueException`, so both are ok.

eis_toolkit/vector_processing/simple_idw.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+@beartype
+def _simple_idw(
+    geodataframe: gpd.GeoDataFrame,
+    target_column: str,
+    resolution: Tuple[Number, Number],
+    extent: Optional[Tuple[float, float, float, float]] = None,
+    power: Optional[Number] = 2,
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
nmaarnio: `Power` should not be declared as an `Optional` here. Optional means that the value can be either None or some type (Number here), but here only a default value is given which is enough. So just change `power: Optional[Number] = 2` to `power: Number = 2`

eis_toolkit/vector_processing/simple_idw.py
```diff
@@ -0,0 +1,113 @@
...
+from eis_toolkit.exceptions import EmptyDataFrameException, InvalidParameterValueException
+
+
+@beartype
+def _simple_idw(
+    geodataframe: gpd.GeoDataFrame,
+    target_column: str,
+    resolution: Tuple[Number, Number],
+    power: Number,
+    extent: Union[Tuple[Number, Number, Number, Number], None],
```
msorvoja, line 17: Although this is equivalent with `Optional[Tuple[Number, Number, Number, Number]]` in simple_idw, I'd declare the parameter type the same way in both functions for consistency.
```suggestion
    extent: Optional[Tuple[Number, Number, Number, Number]]
```

eis_toolkit/vector_processing/simple_idw.py
```diff
@@ -0,0 +1,113 @@
...
+
+def _idw(x, y, z, xi, yi: np.ndarray, power: Number) -> np.ndarray:
+    dist = _distance_matrix(x, y, xi, yi)
+    # Add a small epsilon to avoid division by zero
+    dist = np.where(dist == 0, 1e-12, dist)
+    weights = 1.0 / (dist**power)
+    weights /= weights.sum(axis=0)
+    interpolated_values = np.dot(weights.T, z)
+
+    return interpolated_values
```
msorvoja, line 71: Perhaps move the computations done in `_distance_matrix` into `_idw` and rename it something like `_simple_idw_core`.
```suggestion
def _simple_idw_core(x, y, z, xi, yi: np.ndarray, power: Number) -> np.ndarray:
    d0 = np.subtract.outer(x, xi)
    d1 = np.subtract.outer(y, yi)
    dist = np.hypot(d0, d1)

    # Add a small epsilon to avoid division by zero
    dist = np.where(dist == 0, 1e-12, dist)
    weights = 1.0 / (dist**power)
    weights /= weights.sum(axis=0)
    interpolated_values = np.dot(weights.T, z)

    return interpolated_values
```

eis_toolkit/vector_processing/simple_idw.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def simple_idw(
+    geodataframe: gpd.GeoDataFrame,
+    target_column: str,
+    resolution: Tuple[Number, Number],
+    extent: Optional[Tuple[Number, Number, Number, Number]] = None,
+    power: Number = 2,
+) -> Tuple[float, float, dict]:
```
msorvoja, line 80: The function `_simple_idw` returns `Tuple[np.ndarray, np.ndarray, np.ndarray]` which is immediately returned as `Tuple[float, float, dict]` in `simple_idw`.
```suggestion
) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
```

msorvoja (changes requested):
Hi,
I noticed a couple of things I'd like you to check/fix.

eis_toolkit/vector_processing/simple_idw.py
nmaarnio: Just noticed the `simple_idw` lacks the @beartype decorator which is why this typing mistake was possible.

nmaarnio (commented):
Looks good! I am still not sure why there is a slight difference between the output of this algorithm and QGIS IDW, but as the difference seems small enough, I think we should get this merged and review this code once more at a later point. Before merging, can you remove the unneeded file changes (a change in pca_plot_testing.ipynb and addition of desktop.ini) @Mtk112

nmaarnio (approved):

PR #168, opened 2023-07-12
142 add create constant raster

msmiyels:
Creates a constant raster based on the users input value with three different options:

1. Based on another raster
2. Based on coordinate pair [north, west] with user-defined pixel size
3. Based on coordinates [north, west, east, south] 

Merged it into **three** **core**- and **one** **call**-function. 

The output will be an `integer `type array, if both `constant `and `nodata` values are **integers**.
If only `nodata `is not integer, the metadata will be **floating point** but the array of **integer** **type** (to avoid useless decimal places when storing). In **any other** case, **float64** will be assigned.

Let me know what you think and if you would like to **split** even the call functions into separate functions or keep it as it is (then, probably some code parts would be redundant).

Cheers,
Micha


eis_toolkit/raster_processing/create_constant_raster.py
```diff
@@ -0,0 +1,186 @@
...
+                        coord_north,
+                        target_epsg,
+                        target_pixel_size,
+                        raster_width,
+                        raster_height,
+                        nodata_value,
+                    )
+
+        constant_value = cast_scalar_to_int(constant_value)
+        nodata_value = cast_scalar_to_int(out_meta["nodata"])
```
nmaarnio, line 173: This comment relates to just code style, but I would reduce nesting here. The elifs and elses after raised exceptions could be removed. Here is one suggestion, but this is not a necessary modification:
```suggestion
    if template_raster is not None:
        out_array, out_meta = _create_constant_raster_from_template(constant_value, template_raster, nodata_value)

    elif all(coords is not None for coords in [coord_west, coord_east, coord_south, coord_north]):
        if raster_height <= 0 or raster_width <= 0:
            raise InvalidParameterValueException("Invalid raster extent provided.")
        if not check_minmax_position((coord_west, coord_east) or not check_minmax_position((coord_south, coord_north))):
            raise InvalidParameterValueException("Invalid coordinate values provided.")
        
        out_array, out_meta = _create_constant_raster_from_bounds(
            constant_value,
            coord_west,
            coord_north,
            coord_east,
            coord_south,
            target_epsg,
            raster_width,
            raster_height,
            nodata_value,
        )

    elif all(coords is not None for coords in [coord_west, coord_north]) and all(
        coords is None for coords in [coord_east, coord_south]
    ):
        if raster_height <= 0 or raster_width <= 0:
            raise InvalidParameterValueException("Invalid raster extent provided.")
        if target_pixel_size <= 0:
            raise InvalidParameterValueException("Invalid pixel size.")

        out_array, out_meta = _create_constant_raster_from_origin(
            constant_value,
            coord_west,
            coord_north,
            target_epsg,
            target_pixel_size,
            raster_width,
            raster_height,
            nodata_value,
        )

    else:
        raise InvalidParameterValueException("Suitable parameter values were not provided for any of the 3 methods.")

    constant_value = cast_scalar_to_int(constant_value)
    nodata_value = cast_scalar_to_int(out_meta["nodata"])
```

nmaarnio (commented):
This looks good overall. I am only wondering if the utility functions are strictly necessary, but if you have found out that they are needed or useful, I'll trust you. The public function is quite parameterized since it allows three modes with different parameter requirements, but I think having one interface for this functionality might be still the best option.

nmaarnio (approved):
Hi, looks good now! I resolved conflicts and made a tiny lint fix, I will merge now.

PR #169, opened 2023-07-17
137 add kriging interpolation

msorvoja:
Adds ordinary 2D kriging interpolation into toolkit.


Mtk112 (commented):
Looks great! I am wondering if the variogram_model should be added as optional parameter?
Additionally I added few suggestions.

msorvoja:
> Looks great! I am wondering if the variogram_model should be added as optional parameter? Additionally I added few suggestions.

I did some testing on different variogram models and decided to include them as optional parameter.

Mtk112 (commented):
Looks good to me!

eis_toolkit/vector_processing/kriging_interpolation.py
```diff
@@ -0,0 +1,106 @@
...
+    extent: Tuple[Number, Number, Number, Number],
+    variogram_model: str,
+    method: str,
+    drift_terms: list,
+) -> Tuple[np.ndarray, dict]:
+
+    coordinates = np.array(list(data.geometry.apply(lambda geom: [geom.x, geom.y, geom.z])))
+    x = coordinates[:, 0]
+    y = coordinates[:, 1]
+    z = coordinates[:, 2]
```
nmaarnio, line 30: I am not sure how often the input data has the interpolation attribute as z-geometry. I think it's more likely that the interpolation attribute is a separate feature, as a column in the geodataframe. I suggest that you either change the function to take a column name for the interpolation or do that and additionally allow for z-geometries to be used for interpolation (this could be a boolean parameter for example).

eis_toolkit/vector_processing/kriging_interpolation.py
```diff
@@ -0,0 +1,106 @@
...
+    return z_interpolated, out_meta
+
+
+@beartype
+def kriging(
+    data: gpd.GeoDataFrame,
+    resolution: Tuple[Number, Number],
+    extent: Tuple[Number, Number, Number, Number],
+    variogram_model: str = "linear",
+    method: str = "ordinary",
```
nmaarnio, line 56: These 2 parameters could be changed to use Literals. This way you could get rid of the separate checks below (beartype would check these instead). You can still test these parameter violations in tests, just change the expected exception to beartype's exception there.
```suggestion
    variogram_model: Literal["linear", "power", "gaussian", "spherical", "exponential", "hole-effect"] = "linear",
    method: Literal["ordinary", "universal"] = "ordinary",
```

nmaarnio (commented):
Looks good overall! Just one idea about changing z-geometry to a user-specified attribute for interpolation and some parameter type modification ideas.

msorvoja:
> Looks good overall! Just one idea about changing z-geometry to a user-specified attribute for interpolation and some parameter type modification ideas.

Pushed suggested changes to GH.

nmaarnio (approved):
LGTM!

PR #172, opened 2023-08-04
99 create surface derivatives

msmiyels:
Draft for **slope** and **aspect** computation in a 3x3 window.
Also added a functionality for **classifiying** the **aspect** into 8 or 16 classes (users's choice).

Also provided a small notebook with some example runs.

### Conversions

Can be integrated as an additional option in the GUI/Toolkit for user access.
- `convert_rad_to_deg()`
- `convert_deg_to_rise()`
- `convert_deg_to_rad()`
- `convert_rise_to_deg()`
- `_convert_rad_to_rise()` this is private since should not be accessed by users or the GUI

These functions are the same as in Pull Request for #171 and included to provide full functionality.

### Partial Derivatives
Needed as basic calculations for every kind of surface derivations included into EIS.
These are based on 3x3 window sizes. However, the code is structured in a way so that other methods or modifications (like 5x5 windows or other basic parameter set from other authors) could be added quite easily.

- `method_horn()`
- `method_zevenbergen()`

### Surface Parameters

- `get_slope()` and `_get_slope()`
- `get_aspect()` and `_get_aspect()`
    - `_mask_aspect()`
    - `_classify_aspect()`

Partial derivatives are the preparation for the following **curvatuve** calculations.

Please check this approach so far, I would then continue to integrate the **curvature** sets.


nmaarnio:
Hi @msmiyels and sorry for delayed review on this one. I could review this very soon, but I wonder if something went wrong with this PR as there are so many conflicts for files that do not concern the features of this PR? Would it be possible for you to either merge the master branch so that only relevant modifications are left or if that will get too complicated, create a new branch from the current master branch where you add only the the relevant modifications and additions?

nmaarnio:
Yes, I think it would be easiest if you set up a new branch that is freshly made from master. Thanks!

PR #173, opened 2023-08-09
149 add statistical tests

msorvoja:
Adds the following computations and statistical hypothesis tests:

- Compute correlation and covariance matrices
- Test normality: Shapiro-Wilk and Anderson-Darling tests
- Test independence: Chi-square test

It is assumed that missing values are handled before the function call and that any variable whose data type isn't float is a categorical variable.


eis_toolkit/statistical_analyses/statistical_testing.py
```diff
@@ -0,0 +1,85 @@
...
+
+def _statistical_tests(
+    data: pd.DataFrame, target_column: str, method: Literal, min_periods: Optional[int], delta_degrees_of_freedom: int
+) -> dict:
+
+    statistics = {}
+    normality = {}
+
+    for column in data.columns:
+        if data[column].dtype != float:  # Categorical variables
```
nmaarnio, line 17: I don't think we can assume that this criterion works for all cases.

eis_toolkit/statistical_analyses/statistical_testing.py
```diff
@@ -0,0 +1,85 @@
...
+        statistics["covariance matrix"] = numerical_columns.cov(min_periods=min_periods, ddof=delta_degrees_of_freedom)
+
+    return statistics
+
+
+@beartype
+def statistical_tests(
+    data: pd.DataFrame,
+    target_column: str,
+    method: Literal["pearson", "kendall", "spearman"] = "pearson",
```
nmaarnio, line 45: This could be "correlation_method"

eis_toolkit/statistical_analyses/statistical_testing.py
```diff
@@ -0,0 +1,85 @@
...
+        statistics["correlation matrix"] = numerical_columns.corr(method=method, min_periods=min_periods)
+        statistics["covariance matrix"] = numerical_columns.cov(min_periods=min_periods, ddof=delta_degrees_of_freedom)
+
+    return statistics
+
+
+@beartype
+def statistical_tests(
+    data: pd.DataFrame,
+    target_column: str,
```
nmaarnio, line 44: I thought about the inputs for a while, and I think it could be better if the inputs were simply two arrays of same size (Numpy array or other sequence). If there are multiple pairings the user wishes to test, they could call this function multiple times.

nmaarnio (commented):
Remember to add the .md file. My comments mainly concern input types as I am thinking in which way this function would the most convenient to use. Feel free to disagree.

eis_toolkit/statistical_analyses/statistical_testing.py
msorvoja: You're correct. The problem is that integers can be both categorical and numerical variables. One way to solve this would be to give categorical and numerical data as two separate inputs. But then the data must somehow be split before the function call.
Another option is that the user specifies which variables (=columns in df) are considered categorical. Any thoughts?

eis_toolkit/statistical_analyses/statistical_testing.py
msorvoja: Sorry, I don't quite follow. What would the other array be for?

eis_toolkit/statistical_analyses/statistical_tests.py
```diff
@@ -0,0 +1,123 @@
...
+    """Check if the input dataframe is empty.
+
+    Args:
+        data: Input DataFrame
+
+    Raises:
+        EmptyDataFrameException: The input DataFrame is empty.
+    """
+    if data.empty:
+        raise exceptions.EmptyDataFrameException("The input DataFrame is empty.")
```
nmaarnio, line 20: This could be moved to _checks/dataframe_folder for reusability.

eis_toolkit/statistical_analyses/statistical_tests.py
```diff
@@ -0,0 +1,123 @@
...
+
+    Args:
+        data: DataFrame containing the input data.
+        target_column: Variable against which independence of other variables is tested.
+
+    Raises:
+        InvalidParameterValueException: The target_column is not in input DataFrame.
+
+    Returns:
+        Test statistics for each variable (except target_column).
```
nmaarnio, line 35: Assumptions of chi2 could be listed in the docstring. Implementation looks overall good, but rather than listing all the results (which can be many if a DF has many columns), the results could be put into a dictionary where the key can detail which colum was used. Also, thinking how this might be called from the plugin, maybe letting the user the specify the included columns is a good idea (as many as they want, and if they leave it None/empty, all columns are included).

eis_toolkit/statistical_analyses/statistical_tests.py
```diff
@@ -0,0 +1,123 @@
...
+        Test statistics for each variable.
+    """
+    check_empty_dataframe(data)
+
+    statistics = []
+    for column in data.columns:
+        statistic, p_value = shapiro(data[column])
+        statistics.append((statistic, p_value))
+
+    return statistics
```
nmaarnio, line 69: Here also docstring can be a bit more detailed if possible. Also the same as above, so output as dictionary.

eis_toolkit/statistical_analyses/statistical_tests.py
```diff
@@ -0,0 +1,123 @@
...
+
+    Returns:
+        Correlation matrix
+    """
+    check_empty_dataframe(data)
+
+    if correlation_method == "kendall" and min_periods is not None:
+        raise exceptions.InvalidParameterValueException(
+            "The argument min_periods is available only with correlation methods 'pearson' and 'spearman'."
+        )
```
nmaarnio, line 96: Do min_periods need to be defined if method is "kendall"? Just wondering

eis_toolkit/statistical_analyses/statistical_tests.py
```diff
@@ -0,0 +1,123 @@
...
+
+    return data.corr(method=correlation_method, min_periods=min_periods)
+
```
nmaarnio, line 99: Rather bind the output to a variable and return the variable, this way its clearer what is returned.

eis_toolkit/statistical_analyses/statistical_tests.py
```diff
@@ -0,0 +1,123 @@
...
+    Returns:
+        Covariance matrix
+    """
+    check_empty_dataframe(data)
+
+    if delta_degrees_of_freedom < 0:
+        raise exceptions.InvalidParameterValueException("Delta degrees of freedom must be non-negative.")
+
+    return data.cov(min_periods=min_periods, ddof=delta_degrees_of_freedom)
```
nmaarnio, line 123: Same as above.

eis_toolkit/statistical_analyses/statistical_tests.py
```diff
@@ -0,0 +1,123 @@
...
+        data: DataFrame containing the input data.
+        correlation_method: 'pearson', 'kendall', or 'spearman'. Defaults to 'pearson'.
+        min_periods: Minimum number of observations required per pair of columns to have valid result. Optional.
+
+    Raises:
+        InvalidParameterValueException: min_periods argument is used with method 'kendall'.
+
+    Returns:
+        Correlation matrix
+    """
```
nmaarnio, line 90: Also could extend the docstring here from the beginning description and/or what is returned.

nmaarnio (changes requested):

eis_toolkit/statistical_analyses/statistical_tests.py
msorvoja: On the contrary, min_periods must not be defined if method is "kendall".

nmaarnio: There was a typo, I meant if method is not "kendall", but seems this is optional parameter. Also the pandas `corr` function seems to not raise exceptions if method is "kendall" and min_periods arg is given, but perhaps this check is good to have for clarity.

eis_toolkit/statistical_analyses/statistical_tests.py
```diff
@@ -0,0 +1,143 @@
...
+    Args:
+        data: Dataframe containing the input data.
+        min_periods: Minimum number of observations required per pair of columns to have valid result. Optional.
+        delta_degrees_of_freedom: Delta degrees of freedom used for computing covariance matrix. Defaults to 1.
+
+    Raises:
+        InvalidParameterValueException: Provided value for delta_degrees_of_freedom is negative.
+
+    Returns:
+        EmptyDataFrameException: The input Dataframe is empty.
```
nmaarnio, line 132: This line seems to be in wrong category (should be under Raises)

nmaarnio (approved):

PR #177, opened 2023-08-15
176 add missing docs, fix doc

nmaarnio:
Added missing docs for fuzzy overlay, creating constant raster, checking raster grids and plotting parallel coordinates.

Also removed duplicate line from environment.yml.


msorvoja (approved):
Looks good.

PR #182, opened 2023-08-25
added performance estimation model

dipak5053:
added the performance estimation model


nmaarnio:
Closing as feature importance and cross-validation are already included.

PR #187, opened 2023-09-07
Modify function output and extent parameter

msorvoja:
Modified function so that it now outputs a raster and metadata (Tuple[np.ndarray, dict]) and modified extent parameter.


nmaarnio (approved):
Looks good, I also renamed the module and main function to be more descriptive and consistent with other names. Merging now

PR #190, opened 2023-09-13
Modify resampling function

nmaarnio:
Resampling function modified to take target resolution as input instead of upscaling factors. Removed option to set pixels non-square.


msorvoja (approved):
Looks good to me!

PR #192, opened 2023-09-29
191 add SMOTETomek for class balancing

nmaarnio:


nmaarnio:
Merging without a review as the dependency updates should be in master branch ASAP and the feature itself is small (and can be easily modified if needed).

PR #197, opened 2023-10-10
Add reclassify raster

lehtonenp:
This PR for reclassifying raster data functions.


lehtonenp:
@nmaarnio, I noticed that there are conlflicts in the requirements.txt file. Do I take in the master branches version numbers or the this branches version numbers?

nmaarnio:
Hi, the requirements.txt does not need to be fully updated, I am more concerned that files related to Poetry stay in order. But they look good at a glance and seems like the conflicts got resolved already.

lehtonenp:
> Hi, the requirements.txt does not need to be fully updated, I am more concerned that files related to Poetry stay in order. But they look good at a glance and seems like the conflicts got resolved already.

I fixed the merge conflicts by the help of emt and nialov. Now, I am on to fix the failing tests!

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    If bands are not given, all bands are used for classification.
+
+    Args:
+        raster: Raster to be classified.
+        interval_size: The number of units in each interval.
+        bands: Selected bands from multiband raster. Indexing begins from one. Defaults to None.
+
+    Returns:
+        Raster classified with defined intervals and metadata.
+    """
```
msorvoja, line 117: Add Raises.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    if bands is not None:
+        if not isinstance(bands, list):
+            raise InvalidParameterValueException("Expected bands parameter to be a list")
+        elif not all(isinstance(band, int) for band in bands):
+            raise InvalidParameterValueException("Expected bands to be a list of integers")
```
msorvoja, line 122: See above.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+
+    out_image, out_meta = _raster_with_defined_intervals(raster, interval_size, bands)
+
+    return out_image, out_meta
+
+
+def _raster_with_equal_intervals(  # type: ignore[no-any-unimported]
+    raster: rasterio.io.DatasetReader,
+    number_of_intervals: int,
+    bands: Optional[Sequence[int]] = None,
```
msorvoja, line 134:
```suggestion
    bands: Optional[Sequence[int]],
```

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+) -> Tuple[Sequence[np.ndarray], dict]:
+
+    array_of_bands = []
+    out_image: Sequence = []
```
msorvoja, line 138:
```suggestion
    out_image = []
```

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    out_meta = raster.meta.copy()
+
+    if bands is not None:
+        for band in raster.read(bands):
+            array_of_bands.append(band)
+    else:
+        array_of_bands = raster.read()
+        bands = np.arange(0, len(array_of_bands), 1).tolist()
```
msorvoja, line 146: See my comment about np.arange above.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+
+    If bands are not given, all bands are used for classification.
+
+    Args:
+        raster: Raster to be classified.
+        number_of_intervals: The number of intervals.
+        bands: Selected bands from multiband raster. Indexing begins from one. Defaults to None.
+
+    Returns:
+        rasterio.io.DatasetReader: Raster classified with equal intervals.
```
msorvoja, line 174: No need to declare return type here.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    """
```
msorvoja, line 175: Add Raises

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    if bands is not None:
+        if not isinstance(bands, list):
+            raise InvalidParameterValueException("Expected bands parameter to be a list")
+        elif not all(isinstance(band, int) for band in bands):
+            raise InvalidParameterValueException("Expected bands to be a list of integers")
```
msorvoja, line 180: See my comment about sequences above

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+
+    out_image, out_meta = _raster_with_equal_intervals(raster, number_of_intervals, bands)
+
+    return out_image, out_meta
+
+
+def _raster_with_quantiles(  # type: ignore[no-any-unimported]
+    raster: rasterio.io.DatasetReader,
+    number_of_quantiles: int,
+    bands: Optional[Sequence[int]] = None,
```
msorvoja, line 192:
```suggestion
    bands: Optional[Sequence[int]],
```

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+) -> Tuple[Sequence[np.ndarray], dict]:
+
+    array_of_bands = []
+    out_image: Sequence = []
```
msorvoja, line 196:
```suggestion
    out_image = []
```

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    out_meta = raster.meta.copy()
+
+    if bands is not None:
+        for band in raster.read(bands):
+            array_of_bands.append(band)
+    else:
+        array_of_bands = raster.read()
+        bands = np.arange(0, len(array_of_bands), 1).tolist()
```
msorvoja, line 204: See comment about np.arange above.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    If bands are not given, all bands are used for classification.
+
+    Args:
+        raster: Raster to be classified.
+        number_of_quantiles: The number of quantiles.
+        bands: Selected bands from multiband raster. Indexing begins from one. Defaults to None.
+
+    Returns:
+        Raster classified with quantiles and metadata.
+    """
```
msorvoja, line 233: Add Raises.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    if bands is not None:
+        if not isinstance(bands, list):
+            raise InvalidParameterValueException("Expected bands parameter to be a list")
+        elif not all(isinstance(band, int) for band in bands):
+            raise InvalidParameterValueException("Expected bands to be a list of integers")
+        elif len(bands) > raster.count:
+            raise InvalidParameterValueException("The number of bands given exceeds the number of raster's bands")
+    if number_of_classes == 0:
+        raise InvalidParameterValueException("number_of_classes parameter is 0")
```
msorvoja, line 407: I'd change this to something like this if negative values are not accepted:
```suggestion
    if number_of_classes <= 0:
        raise InvalidParameterValueException("The number of classes must be at least one")
```
Feel free to disagree.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    if bands is not None:
+        for band in bands:
+            stddev = np.nanstd(band)
+            mean = np.nanmean(band)
+            band_statistics.append((mean, stddev))
+    else:
+        for band in range(1, raster.count + 1):
+            stddev = np.nanstd(band)
+            mean = np.nanmean(band)
+            band_statistics.append((mean, stddev))
```
msorvoja, line 433: The type of variable band is integer so you are taking np.nanstd and np.nanmean of an integer which probably was not the intention?

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    If bands are not given, all bands are used for classification.
+
+    Args:
+        raster: Raster to be classified.
+        number_of_intervals: The number of intervals.
+        bands: Selected bands from multiband raster. Indexing begins from one. Defaults to None.
+
+    Returns:
+        Raster classified with standard deviation and metadata.
+    """
```
msorvoja, line 473: Add Raises.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,484 @@
...
+    if bands is not None:
+        if not isinstance(bands, list):
+            raise InvalidParameterValueException("Expected bands parameter to be a list")
+        elif not all(isinstance(band, int) for band in bands):
+            raise InvalidParameterValueException("Expected bands to be a list of integers")
```
msorvoja, line 478: See my comment about sequences above.

tests/raster_processing/reclassify_raster_test.py
msorvoja, line 1: You should be testing the actual functions you've implemented to see that they work as expected.

msorvoja (changes requested):
Hi! I left you some comments mostly about style but also spotted some problems in the code. Please let me know if you disagree with something.

lehtonenp:
Thank you for the thorough checking! I'll start fixing the right away.

tests/raster_processing/reclassify_raster_test.py
lehtonenp: @msorvoja, at the moment the tests showcase the very inner functionality of the functions. The tests are the very minimal example to understand what the functions do to data.

lehtonenp:
@nmaarnio I am reaching out to you as I haven't gotten a response from msorvoja. Is this PR ok?

nialov:
I am taking a quick look now.

nialov:
As I understand, the business logic implemented in `reclassify_raster.py` is reimplemented in the test file (`reclassify_raster_test.py`). This is not the correct way to use unit tests as the logic should only be implemented in the actual code (`reclassify_raster.py`) and then that logic is tested through importing that module and calling the functions. The actual code in `reclassify_raster.py` is now not tested, rather the tests just test themselves.

The tests should follow this kind of schema:

~~~python
from eis_toolkit.raster_processing import reclassify_raster

def test_raster_with_manual_breaks():

    # Define test data here or as as module variables like `test_array` at the start of the file
    raster, number_of_classes, bands = ...
    result = reclassify_raster._raster_with_manual_breaks(raster, number_of_classes, bands)

    assert isinstance(result, tuple)
    assert isinstance(result[0], list)
    assert isinstance(result[1], dict)

    # If you have an expected result you can test that the result stays the same
    # (regression test)
    np.testing.assert_allclose(result[0], expected_output)
~~~

Another note is that the private functions all take `raster: rasterio.io.DatasetReader` as arguments:

~~~python
def _raster_with_natural_breaks(  # type: ignore[no-any-unimported]
    raster: rasterio.io.DatasetReader,
    ...
) -> Tuple[Sequence[np.ndarray], dict]:
~~~

To make it simpler to test the functions, you could simplify/create functions to process one band at a time as there is no interaction between the band data, like so:

~~~python
def _raster_with_natural_breaks_band(  # type: ignore[no-any-unimported]
    band:np.ndarray,
    number_of_classes: int,
) -> np.ndarray:

    breaks = mc.JenksCaspall(band, number_of_classes)
    data = np.digitize(band, np.sort(breaks.bins))
    return data

~~~
These are much easier to test in the `pytest` test file.

The tests should be fixed but this other issue is not a problem for merging, I just think it would simplify the testing and code in general.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,511 @@
...
+        for band in raster.read(bands):
+            array_of_bands.append(band)
+    else:
+        array_of_bands = raster.read()
+
+    for i in range(len(array_of_bands)):
+
+        data_array = array_of_bands[i]
+        # Masking missing values with np.nan
+        data_array[data_array == -1.0e32] = np.nan
```
nialov, line 338: Maybe this is some `numpy` thing but why are values of `-1.0e32` equal to `np.nan`? Should this be a function input that can be given by the user?

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,511 @@
...
+
+        median_value = np.nanmedian(data_array)
+        max_value = np.nanmax(data_array)
+        min_value = np.nanmin(data_array)
+
+        data_array = np.ma.masked_where(np.isnan(data_array), data_array)
```
nialov, line 344: Is this line needed? I do not see the array mask being used, instead the array is compared to `np.nan` everywhere anyway instead of using `data_array.mask` to see what is `np.nan`.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,511 @@
...
+        InvalidParameterValueException: Bands is not a list, bands is not a list of integers,
+        bands exceeds the number of bands in the raster.
+    """
+    if bands is not None:
+        if not isinstance(bands, list):
+            raise InvalidParameterValueException("Expected bands parameter to be a list")
+        elif not all(isinstance(band, int) for band in bands):
+            raise InvalidParameterValueException("Expected bands to be a list of integers")
+        elif len(bands) > raster.count:
+            raise InvalidParameterValueException("The number of bands given exceeds the number of raster's bands")
```
nialov, line 507: These checks are repeated across all the public functions. They could be refactored into a function that is then called to check the inputs to increase code reuse. E.g.

~~~python
def _check_reclassify_raster_inputs(bands):
    if not isinstance(bands, list):
        raise InvalidParameterValueException("Expected bands parameter to be a list")
    elif not all(isinstance(band, int) for band in bands):
        raise InvalidParameterValueException("Expected bands to be a list of integers")
    elif len(bands) > raster.count:
        ...
~~~

lehtonenp:
@nialov, I kindly ask you to review the code. I would pay attention to the public functions and tests. In summary I have modified the private functions to calculate only one band per call of the private function. I made loops to the public functions to iterate all the bands or the bands given by the user. So, is the functionality of iterating over all or the given bands a wanted feature?

nialov:
Trying to review tomorrow or at the start of next week!

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,463 @@
...
+
+    out_image = []
+    is_integer = np.issubdtype(band.dtype, int)
+    is_float = np.issubdtype(band.dtype, float)
+
+    if is_integer:
+        mask = band == nan_value
+        band[mask] = nan_value
+    elif is_float:
+        band[band == nan_value] = np.nan
```
nialov, line 304: Why is the logic different for integers and floats? If the `nan` value for integers is a normal integer (e.g. -9999) it will affect the interval determination, will it not? I think `np.nan` should also be set to integer bands unless I am not understanding things correctly.

nialov (changes requested):
Please add tests also to the main functions: Example below:

~~~python
from tests.raster_processing.clip_test import raster_path as SMALL_RASTER_PATH

def test_raster_with_manual_breaks_main():
    with rasterio.open(SMALL_RASTER_PATH) as raster:
        result = reclassify_raster.raster_with_manual_breaks(
                raster=raster,
                breaks=[1, 2, 3],
                bands=[1],
                )
    assert isinstance(result, Sequence)
    assert isinstance(result[0], np.ndarray)

~~~

Currently the main functions are broken as the inputs to the private functions are wrong.

> I made loops to the public functions to iterate all the bands or the bands given by the user. So, is the functionality of iterating over all or the given bands a wanted feature?

Personally I believe functions that do not need the raster metadata and that do not have interaction between the bands, such as this reclassification, should only work on the individual band data, i.e. `numpy` arrays. So the private functions you have defined would fit this purpose already. But currently in `eis_toolkit` the functions usually take the `rasterio.io.DataSetReader` as input and all do the same band reading internally so maybe keep the main functions so the functions are equivalent to others in the toolkit. @nmaarnio can comment on this if you want to!

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,463 @@
...
+        if not isinstance(band, Sequence):
+            raise InvalidParameterValueException("Expected bands parameter to be a list")
+        elif not (isinstance(band, int)):
+            raise InvalidParameterValueException("Expected band to be an integer")
+
+
+def _breaks_exception(breaks: Sequence[int]):
+
+    if not all(isinstance(_break, int) for _break in breaks):
+        raise InvalidParameterValueException("Expected breaks to contain only integers")
```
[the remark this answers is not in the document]
lehtonenp: @nialov do you mean that I should check that the variable named `band` contains negative integers or that the variable `band` is just a negative value?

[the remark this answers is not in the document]
nialov: Bands should be positive integers (0, 1, 2, ...). You can check that each value in `band` is a positive integer. The type check s are not needed.

eis_toolkit/raster_processing/reclassify_raster.py
nialov: Ah yes true. In this case I believe you need to use `numpy.ma.masked_array` to make sure the chosen integer nan value does not affect the computations. 

~~~python
# nan_value is either a set integer (e.g. -9999) or np.nan
mask = band == nan_value
masked_array = np.ma.masked_array(data=band, mask=mask)

median_value = np.ma.median(masked_array)
...
~~~

You do not need to check for `np.nan` values anywhere after this i.e. no `nan_condition`. Just use the same name methods from `np.ma` instead of `np`. E.g. `np.where` -> `np.ma.where`. You should check that the output, however, is not a masked array. A masked array can be transformed back to a normal array with `np.array(masked_array)`. This transformation will add back the earlier defined nan_value to the array.

lehtonenp:
@nialov, I have made the requested changes to the reclassify methods and tests. Is there anything more to add?

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,422 @@
...
+
+def _raster_with_geometrical_intervals(
+    band: np.ndarray, number_of_classes: int, nan_value: Union[int, float]
+) -> np.ndarray:
+
+    # nan_value is either a set integer (e.g. -9999) or np.nan
+    mask = band == nan_value
+    masked_array = np.ma.masked_array(data=band, mask=mask)
+
+    median_value = np.nanmedian(masked_array)
```
nialov, line 271:
```suggestion
    median_value = np.ma.median(masked_array)
```

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,422 @@
...
+        tail_values = masked_array[np.ma.where((masked_array > median_value))]
+        range_tail = max_value - median_value
+        tail_values = tail_values - median_value + range_tail / 1000.0
+    else:  # Small end tail longer
+        tail_values = masked_array[np.ma.where((masked_array < median_value))]
+        range_tail = median_value - min_value
+        tail_values = tail_values - min_value + range_tail / 1000.0
+
+    min_tail = np.nanmin(tail_values)
+    max_tail = np.nanmax(tail_values)
```
nialov, line 288:
```suggestion
    min_tail = np.ma.min(tail_values)
    max_tail = np.ma.max(tail_values)
```

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,422 @@
+import mapclassify as mc
+import numpy as np
+import rasterio
+from beartype import beartype
+from beartype.typing import Sequence, Tuple, Union
+
+from eis_toolkit.exceptions import InvalidParameterValueException
+
+
+def _bands_non_negative(band: Sequence):
```
nialov, line 10:
```suggestion
def _check_bands_non_negative(band: Sequence):
```

nialov (changes requested):
Looks good, only few small things:

-   Remove `tests/data/small_raster.tif.aux.xml` file. It was probably just added by opening the raster in e.g. ArcGIS or QGIS.

After these, LGTM!

lehtonenp:
> Hey, I took a quick look at this and wrote some comments. I would consider renaming this module to "reclassify.py" and the tool names to `reclassify_<strategy>` or `reclassify_with_<strategy>`. Any thoughts about these renamings @lehtonenp and @nialov ?

@nmaarnio, `reclassify_with_<strategy>` is a suitable name. I can change the function naming.

eis_toolkit/raster_processing/reclassify_raster.py
```diff
@@ -0,0 +1,422 @@
...
+    output = np.array(values_out)
+
+    return output
+
+
+@beartype
+def raster_with_geometrical_intervals(  # type: ignore[no-any-unimported]
+    raster: rasterio.io.DatasetReader, number_of_classes: int, nan_value: Union[int, float], bands: Sequence[int]
+) -> Tuple[Sequence[np.ndarray], dict]:
+    """Classify raster with geometrical intervals (Torppa, 2023).
```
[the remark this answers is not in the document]
lehtonenp: It is true that this is not a novel functionality. However, it is different from some other geometrical intervals. That is why I credited Torppa. The credit can be left out.

nmaarnio (approved):
LGTM now!

nialov (approved):
LGTM! Very well coded functionality!

nmaarnio:
I actually encountered a few issues when implementing CLI functions for these tools just now. `check_raster_bands` expects the `bands` parameter to be a Sequence. Now `None` can be delivered to it. Furthermore, I realized that the functions are returning sequence of 2D Numpy arrays, whereas a 3D Numpy array is how mulitband raster data is usually represented (and can be easily saved with Rasterio).

I ended up creating a suggestion for refactored version of the public relcassification functions. Here cases where `bands` is None or empty sequence are handled, 3D Numpy array allocated and returned. Feel free to modify this or disagree, I just crafted it while thinking of these issues!

```
@beartype
def reclassify_with_manual_breaks(  # type: ignore[no-any-unimported]
    raster: rasterio.io.DatasetReader,
    breaks: Sequence[int],
    bands: Optional[Sequence[int]] = None,
) -> Tuple[np.ndarray, dict]:
    """Classify raster with manual breaks.

    If bands are not given, all bands are used for classification.

    Args:
        raster: Raster to be classified.
        breaks: List of break values for the classification.
        bands: Selected bands from multiband raster. Indexing begins from one. Defaults to None.

    Returns:
        Raster classified with manual breaks and metadata.

    Raises:
        InvalidParameterValueException: Bands contain negative values.
    """

    if bands is None or len(bands) == 0:
        bands = range(1, raster.count + 1)
    else:
        check_raster_bands(raster, bands)

    out_image = np.empty((len(bands), raster.height, raster.width))
    out_meta = raster.meta.copy()

    for i, band in enumerate(bands):
        band_data = raster.read(band)
        out_image[i] = _reclassify_with_manual_breaks(band_data, breaks)

    return out_image, out_meta
```

nmaarnio:
Some changes to` poetry.lock` and `pyproject.toml` had slipped to this PR in some merge, I removed them. Looks good for merging now!

PR #204, opened 2023-10-12
112 toolkit plugin interface

nmaarnio:
The CLI using Typer is in progress and an interface function has been created for most of existing tools.


nmaarnio:
The CLI remains unifinished as some tools are not implemented yet. Also, the CLI might have bugs, but the progress of the branch will be merged now to be included in beta testing.

PR #205, opened 2023-10-13
117 add plotting functions

nmaarnio:
Decided to not create wrappers for seaborn but simply place selected plotting functions from the package under an EIS Toolkit module. Added notebooks with plotting examples.


msorvoja (approved):
Looks good to me!

PR #212, opened 2023-10-24
Update dependencies and delete IOCG test data

nmaarnio:
Updated some packages to more recent versions (Conda env dependencies not updated in this PR).

Deleted IOCG test data as it was a bit too large to store in GH.


nmaarnio:
Merging as this PR simply updates dependencies (tested that Poetry env still works) and deleted chunky IOCG data from remote.

PR #217, opened 2023-10-30
216 modify modules

nmaarnio:
The following changes were made:

- Removed `spatial_analyses` (contents were moved to `vector_processing`)
- Removed `statistical_analyses` (contents were moved to `exploratory_analyses`)
- Moved `checks` under `utilities`
- Removed `dummy_test`
- Imports were adjusted for all files where the above modules were used
- Fixed two docstrings
- Added several* missing doc files


Mtk112 (approved):
Went through the changes, looks good to me. Tests ok, notebooks ok, docs ok.
Merging.

PR #226, opened 2023-11-11
208 optimize IDW

RaineEkman:
- Allocate less memory in _idw_core, dependent only on output size and not the number of points
- Pass smaller arrays into _idw_core
- Don't sort the points to interpolate on

### DIFF
--- a/eis_toolkit/vector_processing/idw_interpolation.py
+++ b/eis_toolkit/vector_processing/idw_interpolation.py
@@ -22,10 +22,7 @@ def _idw_interpolation(
     values = geodataframe[target_column].values
 
     if extent is None:
-        x_min = geodataframe.geometry.total_bounds[0]
-        x_max = geodataframe.geometry.total_bounds[2]
-        y_min = geodataframe.geometry.total_bounds[1]
-        y_max = geodataframe.geometry.total_bounds[3]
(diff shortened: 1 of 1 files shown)


nmaarnio (approved):
Thanks! I can verify that this modification indeed is much better implementation of IDW: The memory use is now just a fraction of what it used to be for large datasets and high resolution outputs can be produced even if they can take a long time. Here are some execution results from my tests with a geochemical dataset of ~1000 points:

#### Output resolution 20 x 20 (3351 x 4066 raster)
  - Old implementation: Tries to allocate ~100GB memory and won't start
  - New implementation: Executed in 4m 12s
  - QGIS IDW: Executed in 3m 42s

#### Output resolution 150 x 150 (446 x 542 raster)
  - Old implementation: Executed in 8.8s
  - New implementation: Executed in 3.2s
  - QGIS IDW: Executed in 4.0s

A note for future: I noticed that the new implementation uses only 1 CPU at a time and uses only a little memory. If we want, we could explore options to increase CPU and memory usage to reach even better performance. It seems IDW of QGIS is also using CPU and memory suboptimally, so there's a good chance to beat their implementation.

RaineEkman:
I don't think we need to increase memory usage with a strategy more or less like this: Split the output array into smaller pieces, invoke a thread on each piece and combine afterwards.

PR #230, opened 2023-11-15
228 optimize distance computation

nialov:
Optimized the performance of `distance_computations` and added additional test cases.

The optimization was done by first merging the geometries, to which distances are calculated, and then calculating the distance for each raster cell to the merged geometries. The merging is done with `unary_union`.

Thanks to @AlVella for pointing this possibility out and providing the draft in which this is based on!


nmaarnio (approved):
LGTM! Do you want to still comment this @AlVella , or will I merge?

AlVella (approved):
Looks great to me, got nothing to add.

Issue #218, opened 2023-10-30
Add functions for CoDA


em-t:
Add functions necessary to perform analysis on compositional data:

The following transformations were requested:

1. pairwise logratio transform
2. additive logratio transform (alr)
3. center logratio transform (clr)
4. isometric logratio transform (ilr)
5. pivot coordinates

nmaarnio:
Implemented and merged in PR #244

PR #244, opened 2023-11-28
218 coda transforms

em-t:
Implement the requested logratio transformations for compositional data (issue #218):

- ILR is only partially implemented: it's included only as a single ILR operation (instead of a full set of a given dataset), and there inverse ILR hasn't been implemented. (ILR is waiting on further details).
- Add a decorator/wrapper for checking that the requirements for compositional data are fulfilled.
- Include a notebook demonstrating use.


em-t:
For the purposes of testing that the functionality is correct, [this paper](https://econ-papers.upf.edu/papers/1627.pdf) was a good overall source (and the first source I found with a proper definition for pivot logratio transform), and [this paper](https://www.lyellcollection.org/doi/pdf/10.1144/gsl.sp.2006.264.01.01) has a simple, follow-along example for CLR, ALR and ILR.

eis_toolkit/transformations/coda/alr.py
```diff
@@ -0,0 +1,85 @@
...
+
+def inverse_alr(df: pd.DataFrame, column: str, scale: float = 1.0, idx: int = -1) -> pd.DataFrame:
+    """
+    Perform the inverse transformation for a set of ALR transformed data.
+
+    Args:
+        df: A dataframe of ALR transformed compositional data.
+        column: The name of the redundant column.
+        scale: The value to which each composition should be normalized. Eg., if the composition is expressed
+            as percentages, scale=100.
```
nmaarnio, line 61: What are valid numbers for scale? I would add some input value check if it has any restrictions. Also, you could change typing hint from `float` to `Number`, as then Beartype won't complain if someone uses an integer.

eis_toolkit/transformations/coda/ilr.py
```diff
@@ -0,0 +1,104 @@
...
+    if check_lists_overlap(subcomposition_1, subcomposition_2):
+        raise InvalidCompositionException("The subcompositions overlap.")
+
+    return _single_ilr_transform(df, subcomposition_1, subcomposition_2)
+
+
+@beartype
+def inverse_ilr():
+    """Perform the inverse transformation for a set of ILR transformed data."""
+    raise NotImplementedError()
```
nmaarnio, line 104: Will this be implemented? If not (if it is not a defined operation or something like that), I would leave it out.

eis_toolkit/transformations/coda/pairwise.py
```diff
@@ -0,0 +1,70 @@
...
+    Args:
+        df: The dataframe containing the columns to use in the transformation.
+
+    Returns:
+        A series containing the transformed values.
+
+    Raises:
+        InvalidColumnException: One or both of the input columns are not found in the dataframe.
+        InvalidParameterValueException: The input columns contain at least one zero value.
+    """
```
nmaarnio, line 63: Here `c1` and `c2` are missing from the docstring. Also, I would rename them in a more descriptive fashion since this is a public function.

eis_toolkit/transformations/coda/pairwise.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@beartype
+def _single_pairwise_logratio(x1: float, x2: float) -> np.float64:
+
+    return np.log(x1 / x2)
+
+
+@beartype
+def single_pairwise_logratio(x1: float, x2: float) -> np.float64:
```
nmaarnio, line 16: These could be typed as `Number` too to allow integers. And renamed as "numerator" and "denominator".

eis_toolkit/utilities/aitchison_geometry.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Returns:
+        A series containing the normalized values.
+    """
+    scale = np.float64(np.sum(row)) / sum_value
+    return np.divide(row, scale)
+
+
+@beartype
+def _closure(df: pd.DataFrame, scale: np.float64 = None) -> pd.DataFrame:
```
nmaarnio, line 24: Maybe this function could be called `normalize_simplex` or something like that?

eis_toolkit/utilities/checks/compositional.py
```diff
@@ -0,0 +1,69 @@
...
+def check_in_simplex_sample_space(df: pd.DataFrame, k: np.float64 = None) -> None:
+    """
+    Check that the compositions represented by the data rows belong to a simplex sample space.
+
+    Checks that each compositional data point belongs to the set of positive real numbers.
+    Checks that each composition is normalized to the same value.
+
+    Args:
+        df: The dataframe to check.
+        k: The expected sum of each row. If None, simply checks that the sum of each row is equal.
```
nmaarnio, line 49: I would rename `k` as `expected_sum`.

eis_toolkit/utilities/miscellaneous.py
```diff
@@ -271,3 +271,43 @@ def set_max_precision(data: Optional[np.ndarray] = None) -> int:
...
+        InvalidColumnIndexException: The amount of provided column names exceeds the amount of columns.
+    """
+    if len(colnames) > df.shape[1]:
+        raise InvalidColumnIndexException()
+
+    columns = [col for col in df.columns]
+    names = dict()
+
+    for i in range(len(columns)):
+        names[columns[i]] = colnames[i]
```
nmaarnio, line 311: Here we might run into indexing error with as `i` can get values beyond length of `colnames`. Since you check in the beginning that the user cannot provide more new columns names than what the DF has columns, I would simply change the content of range(len(()) to colnames:

```suggestion
    for i in range(len(colnames)):
        names[columns[i]] = colnames[i]
```

nmaarnio (changes requested):
Hey! Overall code looks great and is easy to follow! My comments mostly concern some naming conventions and suggestion – feel free to disagree or counter-suggest something for these. I tried to check the different logratio logics too, but I have to admit I am not familiar with these methods myself.

eis_toolkit/transformations/coda/ilr.py
em-t: This should be implemented according some excel sheet I found on Teams, so I left it as a memo. But as we still don't even know how the full set of ILR transforms should be produced, might as well remove this as well. I'll clean it up 👍

eis_toolkit/transformations/coda/alr.py
```diff
@@ -18,68 +20,68 @@ def _alr_transform(df: pd.DataFrame, columns: Sequence[str], denominator_column:
...
-def alr_transform(df: pd.DataFrame, idx: int = -1, keep_redundant_column: bool = False) -> pd.DataFrame:
+def alr_transform(df: pd.DataFrame, column: int = -1, keep_denominator_column: bool = False) -> pd.DataFrame:
     """
     Perform an additive logratio transformation on the data.
 
     Args:
         df: A dataframe of compositional data.
-        idx: The integer position based index of the column of the dataframe to be used as denominator.
+        column: The integer position based index of the column of the dataframe to be used as denominator.
             If not provided, the last column will be used.
```
nmaarnio, line 30: Did you leave the type of this parameter `int` on purpose? What I suggested was to let the user give the desired column name (str) instead of the index. This solution isn't bad, but can be a bit tricky if there are tens of columns in a DF :)

nmaarnio (commented):
Thanks for the quick changes – looks really good to me now! I only had one comment/question about the `alr_transform` parameterization.

eis_toolkit/transformations/coda/alr.py
em-t: I must have missed it or forgotten halfway through going through the change suggestions. 😄 It's now fixed!

nmaarnio:
One more thing I noticed: the doc files are missing for these functions. After you have added them I will merge!

em-t:
I added doc files similar to those for other modules. But I couldn't get mkdocs serve or build to work as described in the instructions.

I'm getting (the same issue is with all of them):

```
ERROR   -  mkdocstrings: eis_toolkit.transformations.coda.alr could not be found
ERROR   -  Error reading page 'transformations/coda/alr.md':
ERROR   -  Could not collect 'eis_toolkit.transformations.coda.alr'
```

Is there something that needs to be done in addition to adding the files for them to work?

nmaarnio:
You need to add an empty `__init__.py` file in the coda folder so that `mkdocs` will recognize it. It should work after that.

nmaarnio (approved):
LGTM, merging!

PR #248, opened 2023-11-30
Add calculate geometry

tomiturunen1:
Calculate geometry from gdf input. Length for LineStrings and area for Polygons. For Points return 0. It's assumed that the user knows the unit of measurement, function doesn't take part in that.


eis_toolkit/vector_processing/calculate_geometry.py
```diff
@@ -0,0 +1,37 @@
...
+
+@beartype
+def calculate_geometry(geodataframe: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
+    """Calculate the length or area of the given geometries.
+
+    Args:
+        geodataframe: Geometries to be calculated.
+
+    Returns:
+        calculated_gdf: Geometries and calculated values-
```
nmaarnio, line 15: Maybe a typo here (line ending with "-" instead of ".")

eis_toolkit/vector_processing/calculate_geometry.py
```diff
@@ -0,0 +1,37 @@
...
+
+def _calculate_value(row):
+    geometry_type = row["geometry"].geom_type
+
+    if geometry_type == "Point":
+        return 0
+    elif geometry_type == "LineString":
+        return row["geometry"].length
+    elif geometry_type == "Polygon":
+        return row["geometry"].area
```
nialov, line 37: Any reason not to handle multi-geometries? I.e. "MultiPoint", "MultiLine...", "MultiPolygon"? These are often mixed or used instead of single geometry types in GIS software. For ease of use I would just handle them as well.

tests/vector_processing/test_calculate_geometry.py
```diff
@@ -0,0 +1,42 @@
...
+@pytest.fixture
+def example_geometries():
+    """Test data, Point, LineString and Polygon."""
+    point = Point(10, 10)
+    line = LineString([(0, 0), (0, 150)])
+    polygon = Polygon([(0, 0), (10, 0), (10, 25), (0, 25)])
+
+    data = {"geometry": [point, line, polygon]}
+
+    return gpd.GeoDataFrame(data, geometry="geometry")
```
nialov, line 19:
```suggestion
    return gpd.GeoDataFrame({"geometry": [point, line, polygon]})
```

More concise

nialov (changes requested):
Few comments. I think this should handle multi-geometries seeing as it requires no extra logic, `shapely` handles the calculation with the same functions transparently. Seeing as we are implementing these very simple GIS functions they should be very generally usable.

nmaarnio:
@tomiturunen1 @nialov Do you think this one is ready now? I personally would have liked the output column to have a more descriptive name than "value", but let's just go ahead and merge if you don't see it an issue.

nialov:
The functionality is so simple that I would not take any more time revising. This can be revised according to the needs that will rise during actual testing.

nmaarnio (approved):

PR #255, opened 2023-12-14
99 create surface derivatives re init

msmiyels:
Added surface attributes calculations and tests.

**First** order (partial derivatives possible after Horn, Evans, Evans-Young or Zevenbergen)
- Aspect **+ Aspect classification for 8 or 16 classes**
- Slope

**Second** order (partial derivatives possible after Evans, Evans-Young or Zevenbergen):
- Plan curvature
- Profile curvature
- Profile minimum curvature
- Profile maximum curvature
- Longitudinal curvature
- Cross-sectional curvature
- Rotor
- Gaussian curvature
- General curvature (total)
- Tangential curvature

All derivatives based on 3x3 windows. If we want to extend it for 5x5 and some 3rd order derivatives, we can do so later (easy to integrate).

- Added notebook for basic demonstration


nmaarnio:
Do you have time to check this one @nialov ? I would prefer to focus on the plugin development right now, but can make some time for this too if needed

tests/surface_attributes/classify_aspect_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+parent_dir = Path(__file__).parent
+raster_path_single = parent_dir.joinpath("../data/remote/small_raster.tif")
+raster_path_multi = parent_dir.joinpath("../data/remote/small_raster_multiband.tif")
+raster_path_nonsquared = parent_dir.joinpath("../data/remote/nonsquared_pixelsize_raster.tif")
+
+
+def test_aspect_classification():
+    
+    for num_classes in [8, 16]:
```
nialov, line 21: You should use `pytest.mark.parametrize` to avoid the for-loop in the test as the loop members do not interact with each other. E.g.

~~~python
@pytest.mark.parametrize("num_classes", [8, 16])
def test_aspect_classification(num_classes: int):
    with rasterio.open(raster_path_single) as raster:
        ...
~~~

tests/surface_attributes/first_order_derivatives_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+parent_dir = Path(__file__).parent
+raster_path_single = parent_dir.joinpath("../data/remote/small_raster.tif")
+raster_path_multi = parent_dir.joinpath("../data/remote/small_raster_multiband.tif")
+raster_path_nonsquared = parent_dir.joinpath("../data/remote/nonsquared_pixelsize_raster.tif")
+
+
+def test_first_order():
+    with rasterio.open(raster_path_single) as raster:
+        methods = ["Horn", "Evans", "Young", "Zevenbergen"]
```
nialov, line 21:
```suggestion
@pytest.mark.parametrize("method", ["Horn", "Evans", "Young", "Zevenbergen"])
def test_first_order(method):
    with rasterio.open(raster_path_single) as raster:
    # Remove for-loop later
```

notebooks/testing_surface_attributes.ipynb
```diff
@@ -0,0 +1,551 @@
...
+        "import numpy as np\n",
+        "import os\n",
+        "import rasterio\n",
+        "\n",
+        "import sys\n",
+        "sys.path.append('..')\n",
+        "\n",
+        "from eis_toolkit.surface_attributes.parameters import first_order, second_order_basic_set\n",
+        "from eis_toolkit.surface_attributes.utilities import classify_aspect, nodata_to_nan\n",
+        "from eis_toolkit.utilities.conversions import convert_deg_to_rad, convert_rad_to_deg, convert_deg_to_rise\n",
```
nialov, line 44: The imports here were wrong. Correct are I believe as such:

~~~python
from eis_toolkit.utilities.nodata import nodata_to_nan
from eis_toolkit.surface_attributes.classification import classify_aspect
~~~

nialov (commented):
Few comments, looks good, especially as you provided a notebook where the results are visualized well! 

There is a lot of math involved so I would like there to be some regression tests that check that we get an expected result from each function call. The expected results could be checked with e.g. the results from `gdaldem` as I understand that conducts some of the same calculations that are provided in this pull request.

eis_toolkit/surface_attributes/parameters.py
```diff
@@ -0,0 +1,213 @@
...
+    if scaling_factor <= 0:
+        raise InvalidParameterValueException("Value must be greater than 0.")
+
+    raster_array = raster.read()
+    raster_array = reduce_ndim(raster_array)
+    raster_array = nodata_to_nan(raster_array, nodata_value=raster.nodata)
+    raster_array = scale_raster(raster_array, scaling_factor)
+
+    cellsize = raster.res[0]
+    p, q, *NoneValues = coefficients(raster_array, cellsize, method)
```
nialov, line 107:
```suggestion
    p, q, *_= coefficients(raster_array, cellsize, method)
```

Underscore is used for collecting values that can be discarded.

tests/surface_attributes/classify_aspect_test.py
```diff
@@ -0,0 +1,129 @@
...
+
+parent_dir = Path(__file__).parent
+raster_path_single = parent_dir.joinpath("../data/remote/small_raster.tif")
+raster_path_multi = parent_dir.joinpath("../data/remote/small_raster_multiband.tif")
+raster_path_nonsquared = parent_dir.joinpath("../data/remote/nonsquared_pixelsize_raster.tif")
+
+
+@pytest.mark.parametrize("num_classes", [8, 16])
+def test_aspect_classification(num_classes: int):
+    results_dict = {
```
nialov, line 21: I would move this outside the function definition and with the name `RESULTS_DICT` to signify that it is a module variable.

tests/surface_attributes/first_order_derivatives_test.py
```diff
@@ -0,0 +1,106 @@
...
+@pytest.mark.parametrize("method", ["Horn", "Evans", "Young", "Zevenbergen"])
+def test_first_order(method: str):
+    result_dict = {
+        "Horn": {"A": 200.727, "G": 7.074},
+        "Evans": {"A": 200.961, "G": 7.017},
+        "Young": {"A": 200.961, "G": 7.017},
+        "Zevenbergen": {"A": 200.929, "G": 7.282},
+    }
+
+    result_dict_min_slope = {"Evans": {"A": 0.396}}
```
nialov, line 28: I would move outside the function, see above comment.

nialov (commented):
I would move the expected value definitions outside the test functions with names like `ASPECT_CLASSIFICATION_RESULTS_DICT` or they could be included in the `pytest.mark.parametrize`. This is just done to reduce the amount of code in the test functions. After these minor changes LGTM.

nialov:
You should merge the master branch into this with `git merge master` and push it. I do not know exactly why the GitHub action checks are not running, maybe that will fix it.

nmaarnio:
Hi @msmiyels and @nialov , is this one ready to be merged in your opinion? My only comment is that there hasn't been a folder called **surface_attributes** in our plans, and that the surface derivatives have been placed under **raster_processing** there. Do you think it would make sense the move them there?

nialov:
Ready to be merged! I will leave the directory question to @nmaarnio. As I see you could do e.g. one of the following:

1. Keep the code in `surface_attributes` folder
2. Create a subdirectory in `raster_processing` like Michael suggested
3. Merge all the code into one or two Python modules (.py files), e.g. `surface_attributes.py` and `surface_attributes_utilities.py` that are put into the `raster_processing` folder

I suppose option 3 would keep the `eis_toolkit` structure the most cohesive.

nmaarnio:
Yes, I would agree with @nialov that the option 3 might be the best. However, you can go for option 2 if you prefer that @msmiyels . Just tag me when you are done and I'll approve and merge then

nmaarnio (approved):
Great, thanks! Merging

PR #258, opened 2023-12-19
Update fuzzy_overlay.py

RichardScottOZ:
minor edits

### DIFF
--- a/eis_toolkit/prediction/fuzzy_overlay.py
+++ b/eis_toolkit/prediction/fuzzy_overlay.py
@@ -94,8 +94,8 @@ def gamma_overlay(data: np.ndarray, gamma: float) -> np.ndarray:
     Args:
         data: The input data as a 3D Numpy array. Each 2D array represents a raster band.
             Data points should be in the range [0, 1].
-        gamma: The gamma parameter. With gamma value 0, result will be same as 'product'overlay.
-            When gamma is closer to 1, the weight of 'sum' overlay is increased.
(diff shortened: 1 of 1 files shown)


nmaarnio:
Hey, thanks for spotting these :+1: I'll go ahead and merge

nmaarnio (approved):

RichardScottOZ (commented):
Ok, why the change from
```python
sum = sum_overlay(data=data)
product = product_overlay(data=data)
return product ** (1 - gamma) * sum**gamma
```
to

```python
product = np.prod(data, axis=0)
sum = data.sum(axis=0) - product
```
```python
return product ** (1 - gamma) * sum**gamma
```

nmaarnio:
> Ok, why the change from
> 
> ```python
> sum = sum_overlay(data=data)
> product = product_overlay(data=data)
> return product ** (1 - gamma) * sum**gamma
> ```
> 
> to
> 
> ```python
> product = np.prod(data, axis=0)
> sum = data.sum(axis=0) - product
> ```
> 
> ```python
> return product ** (1 - gamma) * sum**gamma
> ```

I don't recall exactly, but now that I quickly checked the code, probably just so that `_prepare_data_for_fuzzy_overlay` wouldn't be run 3 times. Not a necessary change, but then again the functions are simple enough at the moment to be maintainable like this.

RichardScottOZ:
Right, that makes sense!

PR #260, opened 2023-12-20
Add findpeaks package

nmaarnio:
Added findpeaks to Poetry dependencies.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -21,5 +21,6 @@ dependencies:
   - rtree >= 1.0.1
   - typer >=0.9.0
   - imbalanced-learn >= 0.11.0
+  - findpeaks >= 2.6.0
   # Dependencies for testing
   - pytest >=7.2.1
--- a/poetry.lock
+++ b/poetry.lock
@@ -467,6 +467,24 @@ files = [
 ]
 
 [[package]]
+name = "caerus"
+version = "0.1.9"
+description = "Python package caerus"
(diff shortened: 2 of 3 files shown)


nmaarnio:
Should have checked this before making the PR, but apparently `findpeaks` is not available in a Conda channel...  What do you think we should do with this @nialov and @msmiyels ? Micha suggested to use this package since it includes some advanced smoothing functionalities, but if I understood correctly, they are not a "must-have"

nialov:
You can add pure `pip` dependencies like so but they are more prone to breaking and there might be installation difficulties in the future.

~~~
diff --git a/environment.yml b/environment.yml
index 0bdc631..aba6582 100755
--- a/environment.yml
+++ b/environment.yml
@@ -22,5 +22,7 @@ dependencies:
   - typer >=0.9.0
   - imbalanced-learn >= 0.11.0
   - findpeaks >= 2.6.0
+  - pip:
+    - findpeaks >=2.6.0
   # Dependencies for testing
   - pytest >=7.2.1
~~~

nmaarnio:
Well, at least the checks pass now. Let's go with this then and solve issues if/when they arise?

nialov:
Yes, if most dependencies are already installed then the `pip` install should not be that prone to breaking.

nmaarnio:
Closing as `findpeaks` was not needed for the smoothing functions

PR #271, opened 2024-01-05
263 add categorical data check

Mtk112:
Added validation check for DataFrame for categorical columns. Created tests for it.
Added the check categorical columns for chi-square test, and tests for it. Adjusted parameters.
Added NonCategoricalDataException.


nmaarnio:
Closing as this is not a necessary inclusion right now. We can reopen when a need occurs.

PR #277, opened 2024-01-10
Added Coda functions to CLI

nmaarnio:
Added:

- ALR transform
- Inverse ALR transform
- CLR transform
- Inverse CLR transform
- Single ILR transform
- Pairwise logratio transform
- Single PLR transform
- PLR transform


msorvoja (approved):
Looks good!

PR #281, opened 2024-01-11
270 change check raster grids to take in raster profiles instead of raster data

msorvoja:
Function `check_raster_grids` in `eis_toolkit/utilities/checks/raster.py` takes in a sequence of raster profiles instead of raster data now. Tests and some dependent functions were modified, see list edited modules below.

Edited modules:
```
eis_toolkit/utilities/checks/raster.py
tests/checks/check_raster_grids_test.py
eis_toolkit/raster_processing/unique_combinations.py
eis_toolkit/utilities/checks/crs.py
eis_toolkit/cli.py
```
Deleted modules (`check_raster_grids.py` was a duplicate):
```
docs/raster_processing/check_raster_grids.md
eis_toolkit/raster_processing/check_raster_grids.py
```


eis_toolkit/utilities/checks/crs.py
nmaarnio, line 1: I would keep the original implementation unless it didn't work. Raster profile might be missing a CRS in some cases, and this implementation will result in error in that case.

We could also move this test into the raster.py file under checks, as it's only used for rasters (and delete this crs.py file).

eis_toolkit/utilities/checks/raster.py
nmaarnio, line 1: Let's allow also raster metadata for the profile parameters since we are using that also a lot in the toolkit. So in practice, just means modifying typing hints from rasterio.profiles.Profile to Union[rasterio.profiles.Profile, dict]. We can still call the parameters raster profiles.

nmaarnio (changes requested):
Just a few small things, otherwise looking good!

eis_toolkit/utilities/checks/crs.py
msorvoja: @nmaarnio Since raster profile (or metadata) is a dictionary, the dot notation won't work (`object.crs`).

msorvoja: How about this?
`@beartype
def check_matching_crs(objects: Iterable) -> bool:
    """Check if every object in a list has a CRS, and that they match.

    Args:
        objects: A list of objects to check.

    Returns:
        True if everything matches, False if not.
    """
    epsg_list = []

    for object in objects:
        if not type(object) == (rasterio.profiles.Profile or dict):
            if not object.crs:
                return False
            epsg = object.crs.to_epsg()
            epsg_list.append(epsg)
        else:
            if "crs" in object:
                epsg_list.append(object["crs"])
            else:
                return False

    if len(set(epsg_list)) != 1:
        return False

    return True`

msorvoja: Now it returns False if object is a profile or dict and it does not contain key "crs".

nmaarnio (approved):
Looks good now!

PR #289, opened 2024-01-15
257 add smoothing functions

msmiyels:
Modularized smoothing functions with example notebook.

There is one `flake8` error, that wants` [x : x, y : y]` with no whitespace between the letters (`[x:x, y:y]`). However, once corrected, black re-formats it to the version with whitespaces. May give an error in the GitHub actions. 

How we want to handle this?

Cheers,
Micha


nmaarnio:
I will review this today or tomorrow! When it comes to the linting issue, add `# noqa: E203` at the end of the problematic line. That should do the trick

nmaarnio (changes requested):
Hey, looks really good overall and the notebook is very illustrative! I noticed a few things regarding the style and docs that I pointed out in the comments.

One a bit bigger question: I noticed that the filters don't seem to necessarily require raster data, but actually any 2D Numpy data (so in other words, raster metadata is not needed). Do you think these functions still best fit under the raster processing category? Maybe more specifically I am thinking about the function signatures, so I propose we either do one of these modifications:
1. Change raster inputs to Numpy array inputs
2. Return the raster metadata in the (out_image, out_meta) format as other raster processing functions, even if the metadata is not modified

Additionally, please also include the `.md` files in the `docs` folder ;). I noticed these are missing for the surface derivatives too, so you can add both in this PR.

nmaarnio:
Looks great now! ✨ It's basically ready to be merged, but I noticed one cell in the notebook didn't unpack the `out_image, out_meta `tuple and therefore threw an error; if you quickly fix that, I'll merge then!

We have had the doc files since the beginning, but I have forgotten them from time to time myself and they haven't been discussed very frequently. However, it was only a couple months ago we started automatically building a PDF out of the docs and also publishing them here: https://gispocoding.github.io/eis_toolkit/

Regarding `findpeaks`: Good, makes things a bit simpler :D

PR #291, opened 2024-01-17
290 add/fix unify rasters CLI function

nmaarnio:


msorvoja (approved):
Looks good and seems to be working as expected.

PR #305, opened 2024-01-30
287 add local morans i

tomiturunen1:
Added Local Moran's I function


tests/exploratory_analyses/local_morans_i_test.py
nmaarnio, line 1: If I recall correctly, we haven't made unit tests so far like they have been implemented here. Meaning that the target values are calculated during the test using the underlying library functions. You know more about testing @nialov , could you comment this?

notebooks/testing_local_morans_i.ipynb
nmaarnio, line 1: Good that you added a notebook for example! However, I think we could use here our geochemical example data that is now in GitHub in the master branch. I think it's preferable to demonstrate using our geological data if convenient and possible, and I believe one of the primary uses for this Local Moran's I might be to inspect spatial autocorrelation of point geochemical data.

So the dataset is called "IOCG_CLB_Till_Geochem_reg_511p.gpkg" and should be found directly under the remote folder. For the target column, any of the mineral columns will probably do.

nmaarnio (changes requested):
Hey, even if I put quite many comments, looks solid overall! Most of my comments relate to cosmetic stuff and are straightforward changes

notebooks/testing_local_morans_i.ipynb
lehtonenp: The notebook's data is "geographical" as I was not comfortable with using example data I do not fully understand. Hence, I used GDP of African countries to demonstrate the correctness of the Local Moran's I.

nmaarnio: Yeah that's absolutely fine. I have also used similar example datasets for some functions. This improvement just came to my mind, since we added the geochem data recently to master and I have personally used Local Moran's I for similar data exploration in my previous job.

lehtonenp: But in the context of EIS it would be better to use the geochem data. I just do not know what columns are feasible for this analysis.

nmaarnio: Yep. I think you could use for example "Li_ppm_551", so concentration of Lithium.

nmaarnio:
@tomiturunen1 are you still working on this? If not, let me know and I can do the test adjustments and we'll get to merge this.

nialov:
> I guess you have done all the updates now? Looks good to me otherwise, but I still think it could be better to compare results with static values rather than use the underlying library within the test. @nialov do you have opinions regarding this?

Static values yes! The current implementation after the recent changes looks about right.

nmaarnio (approved):
Yep looks fine now, merging!

PR #307, opened 2024-01-31
293 add smoothingfilter cli functions

msorvoja:
Adds cli functions for the filter tools. There is a problem with Lee enhanced, frost and kuan filters which I was unable to solve, hence the draft PR. Could you @msmiyels look into it as you have implemented these tools? Those three filter function all give `SystemError: <built-in function len> returned a result with an exception set` with the same data (small_raster.tif in `tests/data/remote`) as the tests for these tools use.


msorvoja:
Hello Michael,

Thank you for looking into this! I'm not sure how familiar you are with these CLI functions so here are (hopefully sufficient) instructions to get you started:

You need a virtual environment into which you install `eis_toolkit`. Inside container,
```
poetry build
```
builds the .whl file inside `dist` folder. Activate the venv if not already and install the wheel
```
pip install eis_toolkit-0.1.0-py3-none-any.whl --force-reinstall
```
Note, that every time you create a new CLI function or make any modifications to the existing ones, you need to rebuild the wheel and install it.

Now with
```
eis --help
```
you should be able to see all the cli functions available. To get help with any of the CLI functions, just type
```
eis function-name-here-cli --help
```
To test, e.g., the Frost filter with the default parameters, type
```
eis frost-filter-cli --input-raster small_raster.tif --output-raster test.tif
```
This should give you the same error as it gives to me.

The Conda fails shouldn't be related to this as they seem to happen in other PRs as well. @nmaarnio is looking into it.

-Mika

nmaarnio:
Hi,

If you are using Poetry (or Docker) environment, you don't actually need to create any additional venv. Just run

`poetry run eis frost-filter-cli --input-raster small_raster.tif --output-raster test.tif` etc.

I tried to solve this quickly, but with no success. I could run the notebook without any issues, but the CLI functions give the same error Mika pointed out. It would still probably be worth test this with different Scipy version, although it puzzles me that the functions run in notebooks/scripts with the current version...

nmaarnio (approved):
Thanks a lot Micha! Worked for me, so let's merge these now.

I removed the debug prints and TODO comments Mika had put there before I merge.

PR #308, opened 2024-02-01
278 add surface derivatives cli functions

msorvoja:
Adds cli functions for public functions in `raster_processing/derivatives/classification.py` and r`aster_processing/derivatives/parameters.py`. Calculation of first and second order surface attributes are combined into a single cli function.


nmaarnio (approved):
Looks good and works!

PR #310, opened 2024-02-01
Change to import type classes from beartype.typing

msorvoja:
Changed imports so that all type classes are imported from beartype.typing.


nmaarnio (approved):
LGTM

PR #315, opened 2024-02-01
Fixed check raster grids to match processing alg

nmaarnio:


msorvoja (approved):
Looks good and works!

PR #323, opened 2024-02-12
Modified normality tests

nmaarnio:
Created dedicated tools for Numpy array and DF inputs. Extended array variant to consider arrays with different dimensions and assume raster-like data (2D is a raster band, 3D is multiband raster data).

Separated code to its own module.


msorvoja (approved):
Looks good!

PR #324, opened 2024-02-13
Implement distance_to_anomaly (Closes #209)

### LINKED ISSUE #209: Add distance to anomaly
Distance to anomaly will be a tool that takes in a (continuous) raster and outputs a raster where pixel values describe the euclidean distance to nearest anomaly (defined with a threshold value).

nialov:
Performance issues are a problem which can be solved by using the alternative functionality that uses gdal_proximity function instead of existing distance_computation implementation. The GDAL implementation is orders of magnitude faster.

Draft for now until I clean up the implementation.

@tomironkko made an initial draft of this function which might have been better than the current `shapely` distance calculation that uses the existing `distance_computation` one but more than likely much worse than the `gdal_proximity` method.


nialov:
Ready for review. The main function, `distance_to_anomaly`, works on all platforms while `distance_to_anomaly_gdal` only works on Linux and MacOS. Based on initial tests I thought the `GDAL` utility functions would work on all platforms but found out later they are inconsistent on Windows. The main problem on Windows is the lack of `osgeo_utils` module in the `conda` environment.

Anyway, thought it best to keep the `GDAL` method for later reference. I just marked the corresponding test as an `xfail` when run on Windows. It is orders of magnitude faster and at least somewhat more memory-efficient (filtering to anomalous values is still done in-memory).

If `distance_computation`, which the main function relies on, requires optimization later, you should check out the original branch by @tomironkko here: <https://github.com/GispoCoding/eis_toolkit/compare/master...209-add-distance-to-anomaly>. The "raster" distance math approach is more efficient but it can obviously only handle raster data while the current `distance_computation` handles vector data as well. You could rasterize first and then use raster math. Alternatively the missing `osgeo_utils` could be investigated.

eis_toolkit/raster_processing/distance_to_anomaly.py
```diff
@@ -0,0 +1,200 @@
...
+    The criteria for what is anomalous can be defined as a single number and
+    criteria text of "higher" or "lower". Alternatively, the definition can be
+    a range where values inside (criteria text of "within") or outside are
+    marked as anomalous (criteria text of "outside").
+
+    Args:
+        raster_profile: The raster profile of which the distances
+            to the nearest anomalous value are determined.
+        anomaly_raster: The raster in which the distances
+            to the nearest anomalous value are determined.
```
nmaarnio, line 40: In the parameters, `anomaly_raster_profile` and `anomaly_raster_data` are separate. I suppose you just forgot to update the docs here?

eis_toolkit/raster_processing/distance_to_anomaly.py
```diff
@@ -0,0 +1,200 @@
...
+
+THRESHOLD_CRITERIA_VALUE_TYPE = Union[Tuple[Number, Number], Number]
+THRESHOLD_CRITERIA_TYPE = Literal["lower", "higher", "in_between", "outside"]
+
+
+@beartype
+def distance_to_anomaly(
+    raster_profile: Union[profiles.Profile, dict],
+    anomaly_raster_profile: Union[profiles.Profile, dict],
+    anomaly_raster_data: np.ndarray,
```
nmaarnio, line 25: A question about these: Is `raster_profile` the target/output profile? For me, it was not (and maybe isn't still) completely clear why we need to specify two raster profiles. In what cases `raster_profile` != `anomaly_raster_profile`? If we keep both, the naming/docs could be clarified.

Also, our raster processing functions have taken in Rasterio's DatasetReaders so far and this implementation conflicts that convention. However, I have to say I prefer this way in general, so maybe this does not need to be changed and we can considering refactoring the other tools.

eis_toolkit/raster_processing/distance_to_anomaly.py
```diff
@@ -0,0 +1,200 @@
...
+    threshold_criteria_value: THRESHOLD_CRITERIA_VALUE_TYPE,
+    threshold_criteria: THRESHOLD_CRITERIA_TYPE,
```
nmaarnio, line 27: These input types do not transfer to the docs well. I would rather just put the Union/Literal types directly to the function signature for better user experience and documentation even if I understand they are used repeatedly in this module.

eis_toolkit/raster_processing/distance_to_anomaly.py
```diff
@@ -0,0 +1,200 @@
...
+) -> np.ndarray:
```
nmaarnio, line 28: We have had the convention of always returning raster profile/meta in raster processing tools. I would follow this schema in this tool too.

eis_toolkit/raster_processing/distance_to_anomaly.py
```diff
@@ -0,0 +1,200 @@
...
+        raise InvalidParameterValueException(
+            f"Expected raster_profile to contain integer width and height. {raster_profile}"
+        )
+
+    raster_transform = raster_profile.get("transform")
+
+    if not isinstance(raster_transform, transform.Affine):
+        raise InvalidParameterValueException(
+            f"Expected raster_profile to contain an affine transformation. {raster_profile}"
+        )
```
nmaarnio, line 61: Is it on purpose that we only check `raster_profile` but not `anomaly_raster_profile`?

While reviewing this I got an idea: should we add a check called `check_raster_meta` or `check_raster_profile` that allows selecting (or deselecting) profile/meta elements to check? This check could essentially do what you are doing here for the specified elements and raise the exceptions too. Currently many of our raster processing functions do not check if metadata contain the required elements.

If we add such a check (and maybe even if we didn't), we could also consider a new dedicated exception type such as `InvalidRasterMetaException`.

nmaarnio (changes requested):
Hey, looks really good overall! I wrote a few comments that mainly regard docs and checks. I think we can keep the GDAL version.

em-t:
I did some testing with the non-gdal version using various raster files and using a randomly generated anomaly profile. Everything has looked good so far :+1: I'm wondering, though, if the function should handle nodata values in some more sophisticated way, or is it ok to assume that the user has replaced nodata values with nan for the input raster?

eis_toolkit/raster_processing/distance_to_anomaly.py
nialov: Yep, updating!

eis_toolkit/raster_processing/distance_to_anomaly.py
nialov: With the initial (non-gdal) approach it was possible to calculate the distances for a raster that is not equivalent in profile to the raster in which the anomalies are determined from. So anomaly raster could have e.g. different resolution to the raster in which the distances were calculated to. For simplification sense I have now removed this option, I suppose it is probably not a common use case. Easier to maintain without it 👍

eis_toolkit/raster_processing/distance_to_anomaly.py
nialov: Done 👍

nialov:
> I did some testing with the non-gdal version using various raster files and using a randomly generated anomaly profile. Everything has looked good so far 👍 I'm wondering, though, if the function should handle nodata values in some more sophisticated way, or is it ok to assume that the user has replaced nodata values with nan for the input raster?

Yes, you are correct that if `np.nan` is not the true nodata value, then the nodata values will be left among the possible anomaly values. I have personally tried to advocate for a centralized approach to handling of nodata but if there are no current plans for such approach, I can implement nodata handling in this function as well (@nmaarnio)?

nialov:
Nodata handling implemented and ready for review again. Nodata only affects the definition of anomalous values in `_fits_criteria`. The output will never have nodata values in it.

eis_toolkit/raster_processing/distance_to_anomaly.py
```diff
@@ -0,0 +1,194 @@
...
+) -> np.ndarray:
+
+    criteria_dict = {
+        "lower": lambda anomaly_raster_data: anomaly_raster_data < threshold_criteria_value,
+        "higher": lambda anomaly_raster_data: anomaly_raster_data > threshold_criteria_value,
+        "in_between": lambda anomaly_raster_data: np.where(
+            np.logical_and(anomaly_raster_data > threshold_criteria[0], anomaly_raster_data < threshold_criteria[1])
+        ),
+        "outside": lambda anomaly_raster_data: np.where(
+            np.logical_or(anomaly_raster_data < threshold_criteria[0], anomaly_raster_data > threshold_criteria[1])
```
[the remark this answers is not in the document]
nialov: Thanks, fixed and added tests for each criteria!

nialov:
Thanks for the review, @nmaarnio! Ready for review again 👍

nmaarnio (approved):
LGTM now!

em-t (approved):
Sorry for leaving my review hanging for so many days. I did some more testing and it looks good to me! I left two minor suggestions, but I'm also ok with merging now. Nice job :+1:

eis_toolkit/raster_processing/distance_to_anomaly.py
```diff
@@ -0,0 +1,224 @@
...
+        threshold_criteria_value=threshold_criteria_value,
+        verbose=verbose,
+    )
+
+
+def _fits_criteria(
+    threshold_criteria_value: Union[Tuple[Number, Number], Number],
+    threshold_criteria: Literal["lower", "higher", "in_between", "outside"],
+    anomaly_raster_data: np.ndarray,
+    nodata_value: Optional[Number],
```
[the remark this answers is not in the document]
nialov: Because the function is a private one I think it is best to make sure all inputs are explicitly passed when using the function so no functionality is missed. Using `Optional` does not mean you need to set the default value, it just means that input is allowed to be passed as `None`.

eis_toolkit/raster_processing/distance_to_anomaly.py
```diff
@@ -0,0 +1,224 @@
...
+
+    cols = np.arange(anomaly_raster_data.shape[1])
+    rows = np.arange(anomaly_raster_data.shape[0])
+
+    all_points_by_rows = [
+        row_points(row=row, cols=cols[data_fits_criteria[row]], raster_transform=anomaly_raster_profile["transform"])
+        for row in rows
+    ]
+    all_points = list(chain(*all_points_by_rows))
+    all_points_gdf = gpd.GeoDataFrame(geometry=all_points, crs=anomaly_raster_profile["crs"])
```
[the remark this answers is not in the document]
nialov: Thanks, added a check that the criteria matches at least some of the data!

nialov:
Ready for review/merging 👍 @nmaarnio, @em-t

nmaarnio (approved):

em-t:
Looks good!

nmaarnio:
Merging!

PR #332, opened 2024-02-22
327 improve readme

nmaarnio:


tmiosmauli (approved):
Looks good to me. Let's do another iteration if we get comments from elsewhere.

PR #335, opened 2024-02-23
329 add cli functions for reclassify raster

nmaarnio:


nmaarnio:
I don't remember if you have implemented/reviewed any CLI implementations before @lehtonenp , but I assigned you here since you worked on both the toolkit func and processing algs for reclassification. To test the CLI functionalities, you should look at the code and try to run each tool. To run the CLI tools, type `poetry run eis reclassify-with-manual-breaks-cli --input-raster my_raster.tif --output-raster out_raster.tif --breaks 1 --breaks 3 --breaks 5` for example

lehtonenp:
> I don't remember if you have implemented/reviewed any CLI implementations before @lehtonenp , but I assigned you here since you worked on both the toolkit func and processing algs for reclassification. To test the CLI functionalities, you should look at the code and try to run each tool. To run the CLI tools, type `poetry run eis reclassify-with-manual-breaks-cli --input-raster my_raster.tif --output-raster out_raster.tif --breaks 1 --breaks 3 --breaks 5` for example

@nmaarnio, I tried to run reclassify-with-manual-breaks-cli and reproject-raster-cli but both of the functions exited with the error that small_raster.tif in the remote folder does not exist. Is this a known error? How should I set up my testing environment; now I do `docker attach eis_toolkit` and try to run the functions with `poetry run eis ...`.

nmaarnio:
Just write out the full path, so `tests/data/remote/small_raster.tif` assuming you are in the project root

lehtonenp:
@nmaarnio, I was able to test the functions all of which worked. But I'm still thinking how I could test the Manual breaks with a meaningful `breaks` parameter? Now, I put 1 as the parameter and the function finished. Or is this relevant at the moment?

nmaarnio:
I am not sure what you mean by a meaningful breaks parameter. If the tool works as intended (creates classes based on the manually specified breaks), then there is no problem? Manual breaks is a standard method to reclassify and the question if a tool should be added or not is relevant when implementing the toolkit function; the CLI funtions are just interfaces on top of the toolkit functions.

lehtonenp:
> then there is no problem? Manual breaks is a standard method to reclassify and the ques

I am sorry for being unclear. I just wished to use a list of values as the breaks parameter for Manual breaks. I was unable to do so.

All of the functions work as intended.

lehtonenp (approved):
As I have stated in the previous comments, all the functions work as intended.

Still, I would like to run the Manual breaks functions with a list of values instead of a number.

Approving nonetheless.

lehtonenp:
> Ah, you meant you wanted to run with multiple breaks. It is a bit clunky, you need to do it like this `--breaks 1 --breaks 3 --breaks 5`, so give the list items separately and repeat the option tag

I tested with multiple --breaks parameter now. Works as expected. Approval persists on my side.

PR #339, opened 2024-03-02
Add standardize and normalize

nmaarnio:


nmaarnio:
@em-t or @lehtonenp , would you have time to review this today or tomorrow?

lehtonenp:
> @em-t or @lehtonenp , would you have time to review this? Or perhaps @msmiyels or @nialov if you are not too busy.

@nmaarnio, I got time to review. So, I'll review this one.

lehtonenp (commented):
I ran tests and cli functions successfully.

I was wondering whether we should have private functions for both normalize and standardize just like in other functions. Are private functions needed? I understand that the functions are short. But the somewhat large if else statements would not be so cluttered as now.

nmaarnio:
Initially we implemented private functions for all tools, but this aspect our style guide has loosened to the point where it is not required anymore. I personally started shifting towards implementing one or more private functions only when I felt they were useful / the function had more contents than just checks and simple call to a library function. Additionally, the gh-pages documentation site generated by `mkdocs` shows the source code of the functions and having only one function makes it show the full code of a function. This could be considered a pro, but I think few people will use it and directly open the source code in their code editor when they wish to inspect our implementations.

Here I felt they would not add much and was not sure which part(s) to separate. However, I am open to suggestions and not against refactoring this! Do you have a specific suggestion for diving the code?

lehtonenp:
I see that the private functions would not add much value to normalize and standardize for the given reasons. It is a plus that `mkdocs` reveals the implementation of the algorithms. Let's keep the code as you have implemented it. Approving.

lehtonenp (approved):
Tests and cli functions work as expected. Private function is not required.

Approving.

nmaarnio:
No worries. I was aware that they can be performed with the existing transformations functions, but I thought it a good idea to create public functions that are called `normalization` and `standardization` (right now I think they are a bit "hidden" in the transformation functions). But in retrospect I now realize it's a bit silly to re-implement them – one option would have been to create public functions called `standardize` and `normalize` that simply call `z_score_normalization` and `min_max_scaling` with locked parameter values.

There are some differences with the inputs and handling. I myself prefer that operations are defined for both vector and raster data if applicable for the plugin and CLI, but defining the core operation on data (Numpy array or Pandas DF) if it makes sense. Maybe we should give this a little thought now, should we merge this or not and how to proceed.

nmaarnio:
Hi @msmiyels , I am revisiting this now. So I agree that a good way to implement for example `normalize` is to just create a wrapper around `min-max` and enforce the `(0, 1)` range. Do you think this brings any additional convenience for users?

I guess accepting tabular data or just vector datasets is another thing. I don't know how many people wish the set of transformation tools to accept vector data input in QGIS plugin, but this would mean changes to all of the transformation functions. So at least for now it would make sense to either create a simple higher-level function for `normalize` or don't add it all.

PR #355, opened 2024-03-15
227 weights generalization improvements

### LINKED ISSUE #227: Weights of evidence: Failure of weights generalization
Add a functionality to handle failure of weights generalization in the weights of evidence method.

There will be instances when generalization of classes for the given studentized contrast threshold will fail for the 'weights_of_evidence_calculate_weights' function. In such cases an informative error could be displayed, and the weights calculations results could be provided to the user, so that the user can evaluate the weights and run the weights calculations again with a suitable threshold.

Alternatively, the users could also be provided with an option of user-defined generalization of classes.

em-t:
A suggestion for how generalization could be handled for the cumulative inputs. Closes #227 but still requires some feedback.

If generalization is unsuccessful during weights calculation, return the weights table without the generalized columns (class, weight and std) and inform the user with a warning (instead of raising an exception and terminating the function).

Make a public function for performing classification and weights calculation for cumulative data, which can be called on a weights table to reclassify the data. Include different options for generalization strategies. (The different options that I provided for the generalization may not make sense. I'll remove any that are unnecessary and make any other modifications requested. Feedback from someone familiar with the method is welcome!)

Code-wise, this can probably be cleaned up some more. It's also noteworthy that this PR introduces custom warnings to the toolkit, which may not be something we want.


nmaarnio:
If you have time at some point @chudasama-bijal , it would be great if you could take a look at this. If the Wofe modifications won't advance otherwise, I will take a look sometime in November and see what I can do.

chudasama-bijal:
This looks fine to me. All generalization options can be retained. The choice between warning and exceptions is something to decide based on the general coding principles defined for this toolkit. Otherwise in terms of method, it seems alright.

nmaarnio:
Great, I'll merge this one too then!

nmaarnio (approved):

PR #357, opened 2024-03-23
Add bayesian nn

zelioluca:
Hi Let s start to check the code together!I m not an expert in bayesian NN but I study the original documentation of tit that is here: 
- https://www.tensorflow.org/probability/overview
- https://keras.io/examples/keras_recipes/bayesian_neural_networks/

One important thing: I got trouble with the package version to install. With the current eis env I suggest you to install the 0.22 the newest one crashed. In this PR there are still some sort of commit from old stuff I do not get way! Monday I will copy a paste there some other documentation to explain what this Bayesian truly do :-)


nmaarnio:
Hi @zelioluca , and sorry for late reply! We agreed with @msmiyels  that he will first work on the CNN and after that on the BNN if he has time. The plan is that Micha would finish the development of these 2 tools, so you might not need to do programming work on these anymore :). We'll ask some questions and help if the need arises!

zelioluca:
Hello niko and micha how are you?ok ok qnything you need I am here! 😁

zelioluca:
Hello there how are you? Yes in my opinion we should use the one that is
working. I m not a baysian guy too... I took the code from another guy and
drop into the plug in.

nmaarnio:
Hi, yes I think it is best to close this PR now and you can start with a new branch @msmiyels . We don't need to delete this branch, so if there is time and will in the future, we can of course return to this implementation if we see a need.

nmaarnio:
Closing now

PR #360, opened 2024-04-03
Modify surface derivative CLI function

nmaarnio:
Modified surface derivative CLI function to match processing algorithm in EIS QGIS Plugin.


nmaarnio:
Merging without review as the modification is simple, functionality is tested and the checks pass now.

PR #362, opened 2024-04-08
Add neural network training plots

nmaarnio:


nmaarnio:
Tests are passing and manually tested for the training history dictionary returned by Keras, which is the intended input. Merging.

PR #363, opened 2024-04-08
346 add confusion matrix plot tool

nmaarnio:


nmaarnio:
Test are passing and function manually tested to work for `sklearn.metrics.confusion_matrix` which is the expected input format. Merging.

PR #364, opened 2024-04-09
Modify raster_to_dataframe

nmaarnio:
- Added nodata handling
- Changed add_coordinates to add geographical coordinates instead of raster array indices


Mtk112 (approved):
Looks good to me!

PR #366, opened 2024-04-09
234 add DBSCAN for rasters

nmaarnio:
Note that k-means branch (PR #365 ) was merged to use same array in test and because of many similarities between these 2 implementations.


nmaarnio:
Tests are passing, CLI functions tested manually and outputs look correct in QGIS. Merging.

PR #371, opened 2024-04-16
370 add raster utility tools

nmaarnio:


nmaarnio:
Tests passing and tested all tools manually. Merging

PR #376, opened 2024-04-24
Update rasterize and vector density

nmaarnio:


nmaarnio:
Tested and the refactored implementations work. Tested together with PR https://github.com/GispoCoding/eis_qgis_plugin/pull/150.

Unit tests temporarily disabled.

Merging.

PR #377, opened 2024-04-24
Add CLI function for Distance to anomaly

nmaarnio:


nmaarnio:
Tested together with https://github.com/GispoCoding/eis_qgis_plugin/pull/151 and works. The failing test doesn't relate to this feature. Merging

PR #381, opened 2024-04-28
Fix fuzzy overlay

nmaarnio:


RichardScottOZ (commented):
Yes, good...a minor style thing...sum often a reserved word or function etc... I would differentiate the variable name

RichardScottOZ (commented):
Looks good I think!

PR #382, opened 2024-04-29
354 distance computation criteria

nmaarnio:


nmaarnio:
Tested CLI and together with EIS QGIS Plugin. Merging.

PR #391, opened 2024-05-23
Add CLI function for CBA

nmaarnio:


nmaarnio:
Failing Conda tests do not relate to this feature. Tested together with https://github.com/GispoCoding/eis_qgis_plugin/pull/175 and works. 

The CLI function supports only one input vector file for now.

PR #395, opened 2024-05-27
Fix Unify rasters CLI function

nmaarnio:


nmaarnio:
Manually tested and the tool should now work properly.

PR #400, opened 2024-05-28
Extend conda installation guidace in README.md

nialov:
Many conda users set the channel priority to strict as part of other guides on conda. Consequently, the installation on Windows will fail as it can not install the updated version of tensorflow from the anaconda channel (part of defaults).


nmaarnio:
Thanks, merging

PR #402, opened 2024-06-03
Fix unify raster grids metadata copying

nmaarnio:
Fix unify raster grids copying dtype and other unwanted metadata from base raster. Refactor code.


Mtk112 (approved):
Looks good to me!

PR #419, opened 2024-08-13
418 add mask raster CLI function

nmaarnio:


msorvoja (commented):
While reviewing this PR, I noticed something which went unnoticed in #417 

_unify_raster_grids masking parameter is `Optional[Literal["extents", "extents_and_nodata"]]` while in the outer function masking is `Optional[Literal["extents", "full"]] = "extents"`

nmaarnio:
> While reviewing this PR, I noticed something which went unnoticed in #417
> 
> _unify_raster_grids masking parameter is `Optional[Literal["extents", "extents_and_nodata"]]` while in the outer function masking is `Optional[Literal["extents", "full"]] = "extents"`

PR #420 should fix that

msorvoja (approved):

Issue #430, opened 2024-09-17
Add proximity tool


nmaarnio:
A tool similar to `distance_computation` is wanted that processes the result raster further. This tool would start with running distance computation but further process the resulting raster so that it would have high values close to the geometries of interest. The values would also (have to) be scaled and a maximum distance defined (the distance at which pixels will be 0).

Name could be `proximity_computation` or `proximity_to` (`distance_computation` could be renamed to `distance_to` too).

Inputs:
- geodataframe
- raster profile
- max distance (not optional, unlike in `distance_computation`)
- scaling method (linear, logarithmic, binary? linear being the default one)

nmaarnio:
Closing as this tool has been implemented

PR #432, opened 2024-09-25
proximity tool to convert the 0's and 1's in the polygon

okolekar:
As discussed during the meetings, where it was requested to add a new tool that can convert the 0's and 1's in the polygon has been created.
I have added the new proximity tool to invert the 0's and 1's in the polygon we get from the distance computation.
I have just used only linear interpolation as of now.

image

### DIFF
--- a/eis_toolkit/vector_processing/proximity_computation.py
+++ b/eis_toolkit/vector_processing/proximity_computation.py
@@ -3,7 +3,7 @@ from numbers import Number
 import geopandas as gpd
 import numpy as np
 from beartype import beartype
-from beartype.typing import Literal, Optional, Tuple, Union
+from beartype.typing import Literal, Tuple, Union
 from rasterio import profiles
 
 from eis_toolkit.transformations.linear import _min_max_scaling
@@ -15,19 +15,18 @@ def proximity_computation(
     geodataframe: gpd.GeoDataFrame,
     raster_profile: Union[profiles.Profile, dict],
     maximum_distance: Number,
-    scaling_method: Literal["linear"],
-    scale_range: Optional[Tuple[Number, Number]] = None,
+    scaling_method: Literal["linear"] = "linear",
+    scale_range: Tuple[Number, Number] = (1, 0),
 ) -> np.ndarray:
-    """Compute proximity to the specified polygons based on the chosen scaling method.
+    """Compute proximity to the nearest geometries.
 
     Args:
-         geodataframe: The GeoDataFrame with geometries to determine distance to.
+         geodataframe: The GeoDataFrame with geometries to determine proximity to.
          raster_profile: The raster profile of the raster in which the distances
                          to the nearest geometry are determined.
-         max_distance: The maximum distance in the output array.
-         scaling_method: method of scaling
-                         linear: linear,
-         scaling_range: a tuple of maximum value and minimum value to scale the distances.
+         max_distance: The maximum distance in the output array beyond which proximity is considered 0.
+         scaling_method: Scaling method used to produce the proximity values. Defaults to 'linear'
+         scaling_range: Min and max values used for scaling the proximity values. Defaults to (1,0).
 
     Returns:
          A 2D numpy array with the scaled values.
@@ -37,8 +36,7 @@ def proximity_computation(
          EmptyDataFrameException: The input geodataframe is empty.
 
     """
-    if scaling_method == "linear":
-        out_matrix = _linear_proximity_computation(geodataframe, raster_profile, maximum_distance, scale_range)
+    out_matrix = _linear_proximity_computation(geodataframe, raster_profile, maximum_distance, scale_range)
 
     return out_matrix
 
@@ -50,12 +48,10 @@ def _linear_proximity_computation(
     maximum_distance: Number,
     scaling_range: Tuple[Number, Number],
 ) -> np.ndarray:
-    """Scales the distance values calculated by the distance_computation function between maximum and minimum.
-
-    Uses linear interpolation to calculate the distance from the polygon.
+    """Compute proximity to the nearest geometries.
 
     Args:
-         geodataframe: The GeoDataFrame with geometries to determine distance to.
+         geodataframe: The GeoDataFrame with geometries to determine proximity to.
          raster_profile: The raster profile of the raster in which the distances
                          to the nearest geometry are determined.
          max_distance: The maximum distance in the output array.


nmaarnio:
Hey, a quick review: The tool should not take in an already computed distance array, but instead perform the distance computation within this proximity tool. It is already possible to first run distance computation and then apply whatever transformation/scaling to the output, but now we want convenience (=user runs just one tool to produce the proximity raster). Also, maximum distance parameter should not be optional here, since the output raster does not really make sense if the values scale up to the boundaries (unless the user specifically sets the max distance so large that it reaches the edges).

nmaarnio:
Also see issue #430 which was made for this feature

okolekar:
> Hey, a quick review: The tool should not take in an already computed distance array, but instead perform the distance computation within this proximity tool. It is already possible to first run distance computation and then apply whatever transformation/scaling to the output, but now we want convenience (=user runs just one tool to produce the proximity raster). Also, maximum distance parameter should not be optional here, since the output raster does not really make sense if the values scale up to the boundaries (unless the user specifically sets the max distance so large that it reaches the edges).

Okay, I get it now.

okolekar:
> Also see issue #430 which was made for this feature

So should I make a commit there or should it be here (I am referring to the pull request here should it be done to #430 or here)?

nmaarnio:
No need to rename branch or anything, the issue you based this feature on was a little misleading (should have been closed already probably) and can be closed when this PR is merged. Additionally #430 will be closed when this is merged.

okolekar:
The tool now takes in a GeoDataFrame, raster profile and maximum distance as inputs and performs the inversion by calling distance computation function inside the tool itself. Logarithmic scaling is also introduced as result of previous.

image


The logarithmic scaling results are shown below

image

eis_toolkit/vector_processing/proximity_tool.py
nmaarnio, line 1: Please rename this file to `proximity_computation.py` (and related files accordingly) to follow the naming schema of distance computation.

eis_toolkit/vector_processing/proximity_tool.py
```diff
@@ -0,0 +1,73 @@
+import numpy as np
+import geopandas as gpd
+from numbers import Number
+from beartype import beartype
+from beartype.typing import Union
+from rasterio import profiles, transform
```
nmaarnio, line 6: transform is unused import

eis_toolkit/vector_processing/proximity_tool.py
```diff
@@ -0,0 +1,73 @@
...
+from eis_toolkit.vector_processing.distance_computation import distance_computation
+
+@beartype
+def calculate_proximity(geodataframe: gpd.GeoDataFrame, raster_profile: Union[profiles.Profile, dict], maximum_distance: Number) -> np.ndarray:
```
nmaarnio, line 10: This line, as some other lines, exceeds the allowed limit. If you use pre-commit locally, it will fix/point out these kind of cosmetic issues.

Also, perhaps it is better if the tool is named similarly as `distance_computation` is right now, so `proximity_computation`

eis_toolkit/vector_processing/proximity_tool.py
```diff
@@ -0,0 +1,73 @@
...
+
+    """ Interpolates the distance values calculated by the distance_computation function between 0 and 1.
+        1 denots the value inside the polygon and 0 at the maximum distance.
+        If maximum_distance value is not provided, the program sets this value to the maximum value 
+        in the provided distance matrix.
+        Uses linear interpolation to calculate the distance from the polygon.
```
nmaarnio, line 16: The format of the docstring should follow that of other tools. I believe the line "If maximum_distance value is not provided, the program sets this value to the maximum value " is also outdated as `maximum_distance` is required

eis_toolkit/vector_processing/proximity_tool.py
```diff
@@ -0,0 +1,73 @@
...
+
+    minimum = np.min(out_matrix)
+    difference = maximum_distance - minimum
+    out_matrix = maximum_distance - out_matrix
+    out_matrix = out_matrix/difference
+
+    return out_matrix
+
+@beartype
+def calculate_logarithmic_proximity(geodataframe: gpd.GeoDataFrame, raster_profile: Union[profiles.Profile, dict], maximum_distance: Number) -> np.ndarray:
```
nmaarnio, line 38: I think it would be better if the choice of scaling should be parameterized, not formulated as another public function. So, let's expose only one `proximity_computation` to the users with parameter `scaling`. It could be literal `Literal['linear', 'log']` with linear as default.

eis_toolkit/vector_processing/proximity_tool.py
```diff
@@ -0,0 +1,73 @@
...
+        Returns:
+            A 2D numpy array with the the distance values inverted.
+    """
+
+    out_matrix = distance_computation(geodataframe, raster_profile, maximum_distance)
+
+    minimum = np.min(out_matrix)
+    difference = maximum_distance - minimum
+    out_matrix = maximum_distance - out_matrix
+    out_matrix = out_matrix/difference
```
nmaarnio, line 33: I believe you could utilize `_min_max_scaling` from EIS Toolkit although I am not 100% sure. Please check and if it fits, we could replace this piece of code with a call to that function to have the operation centralized in one place only. 

There's also the point that if we extend the scaling options in the future, it is good to use our own `transformations` module wherever we can

eis_toolkit/vector_processing/proximity_tool.py
```diff
@@ -0,0 +1,73 @@
...
+    minimum = np.min(distance_array)
+    if(minimum != 0):
+        log_minimum = np.log(minimum)
+    else :
+        log_minimum = minimum
+    difference = log_maximum - log_minimum
+    out_matrix = log_maximum - out_matrix
+    out_matrix = out_matrix/difference
+
+    out_matrix = np.nan_to_num(out_matrix,nan=log_maximum)
```
nmaarnio, line 71: Check also if the `transformations` module would be of use to perform the log transform.

nmaarnio (changes requested):
Hi Omkar, did a full review now. Apologies for the strict format rules of EIS Toolkit, but we try to keep the toolkit harmonized and tidy.

Besides the comments I wrote, please add a .md file in the docs folder and unit tests (see some other unit tests in EIS Toolkit for examples, no need to include many).

Additionally, you could add a CLI function in `cli.py` for `proximity_computation`, but I can take care of that after this PR has been merged if you want.

eis_toolkit/vector_processing/proximity_tool.py
okolekar: I had an issue with the pre-commit hooke
[pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available](https://stackoverflow.com/questions/45954528/pip-is-configured-with-locations-that-require-tls-ssl-however-the-ssl-module-in)

However, is now resolved. I ran the precommit hooke and cleared the cosmetic problems
image

eis_toolkit/vector_processing/proximity_tool.py
okolekar: I tried using the _min_max_scaling, using this tool I could not achieve the inversion we intend to achieve. Moreover, I cannot see the scaling happening as well. Below I have attached the pictures.

image

nmaarnio: What parameters did you use for min-max scaling? I noticed that that the tool does not allow `new_range=[(1, 0)]` right now, but you can try commenting out the lines 167-169 in `min_max_scaling.py` and then run it with `new_range=[(1, 0)]`. I think it should produce the correct results, just make sure you apply the max distance cut before the scaling operation.

okolekar: > I believe you could utilize `_min_max_scaling` from EIS Toolkit although I am not 100% sure. Please check and if it fits, we could replace this piece of code with a call to that function to have the operation centralized in one place only.
> 
> There's also the point that if we extend the scaling options in the future, it is good to use our own `transformations` module wherever we can

Hi niko,
as per your previous comment, I have used the '_min_max_scaling' instead of 'min_max_scaling.py.'

nmaarnio: Hi,
Ah yes you're right, I suggested to use `_min_max_scaling` initially – that would work just as well (also using range `(1, 0)`). I do not know why you are getting incorrect results, but if you run the scaling as the last operation, the data should definitely end up in the given range and not 0-25 like in the image you attached. I am able to produce the correct proximity computation result with this method that I suggest. You can push your new implementation to GitHub and I can take a look if you want.

okolekar: Hi,
I did find the problem, there was a naming error in my code, I used the same variable to get maximum distance and for scaling . Due to which the value for scaling was overridden. I found it. Also, just a small query I have here, for log scaling, there are various methods (log to base 10/ natural log), should there be another variable that the user must provide inorder to select the base of the log?

nmaarnio: Hi, ok good! Not another variable I would say, rather you can include the log variations in the scaling parameter like this `Literal['linear', 'ln', 'log2', 'log10']`

nmaarnio:
Any updates with this tool @okolekar? Since the last progress of EIS project is nearing and is less than a month away, it would be good if this tool was implemented and merged soon so that people have the opportunity to test it.

okolekar:
> Any updates with this tool @okolekar? Since the last progress of EIS project is nearing and is less than a month away, it would be good if this tool was implemented and merged soon so that people have the opportunity to test it.

I am very sorry for the delay, as we face a small challenge with the log transformation. However, I am done with the linear tool. The log transformation is a bit more tricky. As the results are not the same as what we expected.
However, if the tool has to be tested I can upload the linear part of the proximity tool after removing the log transformation. Shall I proceed to upload with the linear part?

nmaarnio:
No worries! Okay, I think in this case we should add the log transform in a later PR and proceed with just the linear one for now. Thanks

okolekar:
Hello @nmaarnio,
I have pushed the changes with the recommended changes. Sorry for the delay as I was working on log transformation. I did not make changes to the cli.py (command line interface) seems to take a bit more time. However, due to the urgency of the project I have skipped it in this commit. I hope that is okay. I will now work with the log transformations. However, I am available to make changes if needed.

nmaarnio:
Thanks, I will do a review tomorrow and implement the CLI function 👍

eis_toolkit/vector_processing/proximity_computation.py
```diff
@@ -0,0 +1,76 @@
...
+from eis_toolkit.transformations.linear import _min_max_scaling
+from eis_toolkit.vector_processing.distance_computation import distance_computation
+
+
+@beartype
+def proximity_computation(
+    geodataframe: gpd.GeoDataFrame,
+    raster_profile: Union[profiles.Profile, dict],
+    maximum_distance: Number,
+    scaling_method: Literal["linear"],
```
nmaarnio, line 18: It's good to provide a default value here.
```suggestion
    scaling_method: Literal["linear"] = "linear",
```

eis_toolkit/vector_processing/proximity_computation.py
```diff
@@ -0,0 +1,76 @@
...
+    scale_range: Optional[Tuple[Number, Number]] = None,
```
nmaarnio, line 19: I think this shouldn't be optional since it is needed to scale the results. Also could provide a default value here.
```suggestion
    scale_range: Tuple[Number, Number] = (1, 0),
```

eis_toolkit/vector_processing/proximity_computation.py
```diff
@@ -0,0 +1,76 @@
...
+    """Compute proximity to the specified polygons based on the chosen scaling method.
+
+    Args:
+         geodataframe: The GeoDataFrame with geometries to determine distance to.
+         raster_profile: The raster profile of the raster in which the distances
+                         to the nearest geometry are determined.
+         max_distance: The maximum distance in the output array.
+         scaling_method: method of scaling
+                         linear: linear,
+         scaling_range: a tuple of maximum value and minimum value to scale the distances.
```
nmaarnio, line 30: Here are my suggestions how to modify this docstring.
```suggestion
    """Compute proximity to the nearest geometries.

    Args:
         geodataframe: The GeoDataFrame with geometries to determine proximity to.
         raster_profile: The raster profile of the raster in which the distances to the nearest geometry are determined.
         max_distance: The distance in the output array beyond which proximity is considered 0.
         scaling_method: Scaling method used to produce the proximity values. Defaults to 'linear'.
         scaling_range: Min and max values used for scaling the proximity values. Defaults to (1, 0).
```

eis_toolkit/vector_processing/proximity_computation.py
```diff
@@ -0,0 +1,76 @@
...
+    Returns:
+         A 2D numpy array with the scaled values.
+
+    Raises:
+         NonMatchingCrsException: The input raster profile and geodataframe have mismatching CRS.
+         EmptyDataFrameException: The input geodataframe is empty.
+
+    """
+    if scaling_method == "linear":
+        out_matrix = _linear_proximity_computation(geodataframe, raster_profile, maximum_distance, scale_range)
```
nmaarnio, line 41: Since we now have only one method, perhaps the if-clause and even the whole scaling_method parameter could be removed and added back when we got more methods. Now in theory the `out_matrix` can be undefined in row 43 below if this if-statement wouldn't be true for whatever reason.

eis_toolkit/vector_processing/proximity_computation.py
```diff
@@ -0,0 +1,76 @@
...
+         max_distance: The maximum distance in the output array.
+         scaling_range: a tuple of maximum value in the scaling and minimum value.
+
+    Returns:
+         A 2D numpy array with the linearly scaled values.
+
+    Raises:
+         NonMatchingCrsException: The input raster profile and geodataframe have mismatching CRS.
+         EmptyDataFrameException: The input geodataframe is empty.
+    """
```
nmaarnio, line 70: It's not necessary to write docstring for private functions, but also it doesn't hurt. If you keep this, modify it according to the public function docstring.

nmaarnio (changes requested):
Looking good, just a few tiny things and then we can merge

eis_toolkit/vector_processing/proximity_computation.py
okolekar: This was kept optional as I had the log transformations as well for which this scale was not working hence was optional. But will make it as required input with the suggested default value

nmaarnio (approved):
Looks good now! Merging

PR #436, opened 2024-10-17
435 fix transformations tools

nmaarnio:
- fix bug where transformations tools fail when nodata is not given by user and not found in raster metadata
- allow to use min-max scaling to invert value range (min can be larger than max in `new_range`)


nmaarnio:
Unit tests passing and tools tested to work, merging.

PR #438, opened 2024-10-22
Add proximity computation CLI function

nmaarnio:


msorvoja (approved):
Appears to be working as expected!

PR #451, opened 2024-11-06
Pin scipy to 1.11.4

nmaarnio:


nmaarnio:
The new build should work – at least it worked for me and @Mtk112 . Merging and we'll see if it works for others too

PR #457, opened 2024-11-13
448 coda transformations data closure

jtlait:
I reverted back to @em-t code and I also slightly altered closure check to be either closure of 1 and 100 as other cases should be very rare.


nmaarnio (approved):
LGTM

PR #461, opened 2024-11-15
Fix saving plots when they are also shown immediately in CLI

nmaarnio:


nmaarnio:
Tested and fixes the issue

PR #463, opened 2024-11-25
fix(SMOTETomek): improve documentation and parameter typing

msorvoja:
Added link to SMOTETomek's documentation in the docstring. Also, changed parameter type of 'sampling_strategy' from `Union[float, str, dict]` to `Union[float, Literal, dict]` to make sure that beartype will roar if the given argument is a str and not one of the correct ones.


nmaarnio:
I don't think we can support passing a dictionary well over the CLI. However, I would think that the float input is needed. Could add an additional CLI parameter for that and decide that either the float or literal overrides the other when given (similar as base raster overrides manual raster settings in some other CLI functions).

@msmiyels , could you help with reviewing this?

eis_toolkit/training_data_tools/class_balancing.py
```diff
@@ -11,24 +11,27 @@
 def balance_SMOTETomek(
     X: Union[pd.DataFrame, np.ndarray],
     y: Union[pd.Series, np.ndarray],
-    sampling_strategy: Union[float, str, dict] = "auto",
+    sampling_strategy: Union[float, Literal["minority", "not minority", "not majority", "all", "auto"], dict] = "auto",
     random_state: Optional[int] = None,
 ) -> tuple[Union[pd.DataFrame, np.ndarray], Union[pd.Series, np.ndarray]]:
```
nmaarnio, line 16: Just realized now that this is a more accurate way to type the return values
```suggestion
) -> Union[tuple[pd.DataFrame, pd.Series]], tuple[np.ndarray, np.ndarray]]:
```

eis_toolkit/cli.py
```diff
@@ -3026,6 +3036,49 @@ def gamma_overlay_cli(input_rasters: INPUT_FILES_ARGUMENT, output_raster: OUTPUT
...
+    output_raster: OUTPUT_FILE_OPTION,
+    output_labels: OUTPUT_FILE_OPTION,
+    sampling_strategy_literal: Annotated[SMOTETomekSamplingStrategy, typer.Option()] = SMOTETomekSamplingStrategy.auto,
+    sampling_strategy_float: Optional[float] = None,
+    random_state: Optional[int] = None,
+):
+    """Resample feature data using SMOTETomek.
+
+    Parameter sampling_strategy_float will override sampling_strategy_literal if given.
+    """
```
nmaarnio, line 3056: A small nitpick, if multiline docstring, the """ should be on a blank line
```suggestion
    """
    Resample feature data using SMOTETomek.

    Parameter sampling_strategy_float will override sampling_strategy_literal if given.
    """
```

eis_toolkit/training_data_tools/class_balancing.py
```diff
@@ -11,24 +11,27 @@
...
+    For more information about Imblearn SMOTETomek read the documentation here:
+    https://imbalanced-learn.org/stable/references/generated/imblearn.combine.SMOTETomek.html.
 
     Args:
-        X: The feature matrix (input data as a DataFrame).
-        y: The target labels corresponding to the feature matrix.
+        X: Input feature data to be sampled.
+        y: Target labels corresponding to the input features.
         sampling_strategy: Parameter controlling how to perform the resampling.
             If float, specifies the ratio of samples in minority class to samples of majority class,
```
nmaarnio, line 27: Could mention here that float is only available for binary classification.

eis_toolkit/cli.py
```diff
@@ -3026,6 +3036,49 @@ def gamma_overlay_cli(input_rasters: INPUT_FILES_ARGUMENT, output_raster: OUTPUT
...
+
+# BALANCE SMOTETOMEK
+@app.command()
+def balance_data_cli(
+    input_rasters: INPUT_FILES_ARGUMENT,
+    input_labels: INPUT_FILE_OPTION,
+    output_raster: OUTPUT_FILE_OPTION,
+    output_labels: OUTPUT_FILE_OPTION,
+    sampling_strategy_literal: Annotated[SMOTETomekSamplingStrategy, typer.Option()] = SMOTETomekSamplingStrategy.auto,
+    sampling_strategy_float: Optional[float] = None,
```
nmaarnio, line 3050: We need to come up with more descriptive names for these, I think. For users it will be confusing what is a "sampling strategy float" and they might not know what "literal" means in this context.

eis_toolkit/cli.py
```diff
@@ -3026,6 +3036,49 @@ def gamma_overlay_cli(input_rasters: INPUT_FILES_ARGUMENT, output_raster: OUTPUT
...
 
+# --- TRAINING DATA TOOLS ---
+
+
+# BALANCE SMOTETOMEK
+@app.command()
+def balance_data_cli(
+    input_rasters: INPUT_FILES_ARGUMENT,
+    input_labels: INPUT_FILE_OPTION,
+    output_raster: OUTPUT_FILE_OPTION,
```
nmaarnio, line 3047: I don't think we can return the stacked rasters as one. We have to return individual rasters that correspond to the input rasters, like in the `unify_rasters` CLI function. However, as I was digging into this, I learned that the balancing process changes the data size and spatial integrity of the data (idk if this is a concern). I think we need to think about this more and consult someone who knows about this method in this context.

nmaarnio (changes requested):
I wrote some actionable suggestions, but either I am missing something obvious or I don't fully understand how this method is applied to raster data. @msmiyels , in case you have time, could you comment about the output data size?

eis_toolkit/training_data_tools/class_balancing.py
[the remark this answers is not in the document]
nmaarnio: Hi @msmiyels and thanks for the comments! I'll get back to this later more in depth, but your last point was what we were looking for most. We were not sure if there is some accepted logic to "squeeze in " the extra synthetic data produced by this tool, but now I understood that there is not. I gather that we should not offer this as a separate tool for the user, but rather parameterize the supervised model training tools that could run this under the hood (?)

PR #468, opened 2024-11-28
466 add feature importance cli

### LINKED ISSUE #466: Add feature importance CLI

msorvoja:
Fixes #466

### DIFF
--- a/eis_toolkit/cli.py
+++ b/eis_toolkit/cli.py
@@ -875,37 +875,37 @@ def local_morans_i_cli(
 @app.command()
 def feature_importance_cli(
     model_file: INPUT_FILE_OPTION,
-    x_input: INPUT_FILE_OPTION,
-    y_input: INPUT_FILE_OPTION,
+    input_rasters: INPUT_FILES_ARGUMENT,
+    target_labels: INPUT_FILE_OPTION,
     output_file: OUTPUT_FILE_OPTION,
-    feature_names: List[str],
+    feature_names: Optional[List[str]] = None,
     n_repeats: int = 50,
     random_state: Optional[int] = None,
 ):
     """Evaluate the feature importance of a sklearn classifier or regressor."""
     from eis_toolkit.exploratory_analyses.feature_importance import evaluate_feature_importance
-    from eis_toolkit.prediction.machine_learning_general import load_model
+    from eis_toolkit.prediction.machine_learning_general import load_model, prepare_data_for_ml
 
     typer.echo("Progress: 10%")
 
     model = load_model(model_file)
     typer.echo("Progress: 20%")
 
-    with rasterio.open(x_input) as src:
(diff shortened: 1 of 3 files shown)


eis_toolkit/cli.py
```diff
@@ -871,6 +871,43 @@ def local_morans_i_cli(
...
 
 
+# FEATURE IMPORTANCE
+@app.command()
+def feature_importance_cli(
+    model_file: INPUT_FILE_OPTION,
+    input_rasters: INPUT_FILES_ARGUMENT,
+    target_labels: INPUT_FILE_OPTION,
+    output_file: OUTPUT_FILE_OPTION,
+    feature_names: Optional[List[str]] = None,
```
nmaarnio, line 881: Feature names are not optional per definition of the toolkit function right now. However, since the features are the input rasters given to this CLI function, I would remove this parameter from the CLI function and simply generate the required `feature_names` (so list of input raster names).

eis_toolkit/cli.py
```diff
@@ -871,6 +871,43 @@ def local_morans_i_cli(
...
+    typer.echo("Progress: 30%")
+
+    feature_importance, result = evaluate_feature_importance(model, X, y, feature_names, n_repeats, random_state)
+    typer.echo("Progress: 80%")
+
+    # np.ndarrays are not JSON serializable -> convert to lists
+    result = {key: value.tolist() for key, value in result.items()}
+
+    feature_importance.to_csv(output_file)
+    json_str = json.dumps(result)
```
nmaarnio, line 904: I don't think we need to give plugin users the whole result dictionary. I think that it suffices to give the mean importances matched with the associated feature name, so what is now returned as a DF. If we want to follow the convention that is now used in EIS Toolkit. a dictionary would be the most suitable form. So a dictionary where keys are feature names and values are the mean importances. However, I've been pondering for some time should all the CLI functions be changed to return files instead of dictionaries. Perhaps best of both worlds would be to optionally return files if it is possible to output file optional in QGIS, I don't remember right now

If you do this modification, remember to remove the `output_file` param

nmaarnio (changes requested):
Couple comments, see below.

Besides those, I looked at the toolkit function and a couple things could be changed there as part of this PR:
- Rename x_test -> X and change description, does not need to be testing data. It can be the same exact data that is fed into any Sklearn model
- Rename y_test -> y and change description, does not need to be testing data
- Change n_repeats default to 10? It is 5 in Sklearn, if we want higher maybe 10 is enough
- Add check that `feature_names`  length matches the dimensions of X (X.shape[1] I think?)
- Maybe we can remove the multiplication by 100 for the importance means – I think the range [0,1] where they reside by default is pretty intuitive and used in other phases of MPM / EIS Toolkit in general

nmaarnio (approved):

PR #472, opened 2024-12-02
Added the gdal related exceptions in the try and except block to avoi…

okolekar:
…d multiple import errors.


nmaarnio:
I'm working on a PR that does effectively the same thing as you did here, so I'll be closing this PR now.

PR #475, opened 2024-12-12
Add WofE CLI functions

msorvoja:
This PR adds CLI functions for the two WofE functions.

In the first CLI function (`weights_of_evidence_calculate_weights_cli`), the resulting dataframe is saved into a CSV file and the arrays are saved as rasters. The number of deposit pixels and evidence pixels are only printed to the user.

In the second CLI function (`weights_of_evidence_calculate_responses_cli`), the resulting probability and confidence arrays are saved as rasters.

In the latter CLI function, there are at least two options how to transform the input rasters into an array of dictionaries needed in the corresponding toolkit function: 
- Create dictionaries based on the number of input rasters and assign key-array pairs so that each dictionary has got the same keys, for example W+ and S_W+ (remove any uncommon keys), or
- read rasters into dictionaries based on their file path. This assumes that each set of outputs of `weights_of_evidence_calculate_weights_cli` are in separate folders. As many dictionaries will be created as there are distinct (innermost) folders in the paths of the input rasters. I'm not sure whether this approach will work with the plugin.

I believe the first approach makes more sense, but I'll let you comment which approach you prefer @nmaarnio.

### DIFF
--- a/eis_toolkit/cli.py
+++ b/eis_toolkit/cli.py
@@ -3124,7 +3124,7 @@ def weights_of_evidence_calculate_weights_cli(
 
     typer.echo(f"Number of deposit pixels: {nr_of_deposits}")
     typer.echo(f"Number of all evidence pixels: {nr_of_pixels}")
-    typer.echo(f"Results saved in {output_dir}.")
+    typer.echo(f"Weight calculations completed, rasters and CSV saved to {output_dir}.")
 
 
 @app.command()
@@ -3178,8 +3178,6 @@ def weights_of_evidence_calculate_responses_cli(
         dst.write(confidence_array, 1)
 
     typer.echo("Progress: 100%")
-
-    # typer.echo(f"Results saved in {output_dir}.")
 
 
 # --- TRANSFORMATIONS ---
--- a/eis_toolkit/cli.py
+++ b/eis_toolkit/cli.py
@@ -3210,6 +3210,36 @@ def weights_of_evidence_calculate_responses_cli(
     )
 
 
+@app.command()
+def agterberg_cheng_CI_test_cli(
+    input_posterior_probabilities: INPUT_FILE_OPTION,
+    input_posterior_probabilities_std: INPUT_FILE_OPTION,
+    nr_of_deposits: Annotated[int, typer.Option()],
+):
+    """Perform the conditional independence test presented by Agterberg-Cheng (2002)."""
+    from eis_toolkit.prediction.weights_of_evidence import agterberg_cheng_CI_test
+
+    typer.echo("Progress: 10%")
+
+    with rasterio.open(input_posterior_probabilities) as src:
+        posterior_probabilities = src.read(1)
+
+    with rasterio.open(input_posterior_probabilities_std) as src:
+        posterior_probabilities_std = src.read(1)
+
+    typer.echo("Progress: 25%")
+
+    _, _, _, _, summary = agterberg_cheng_CI_test(
+        posterior_probabilities=posterior_probabilities,
+        posterior_probabilities_std=posterior_probabilities_std,
+        nr_of_deposits=nr_of_deposits,
+    )
+
+    typer.echo("Progress: 100%")
+    typer.echo("Conditional independence test completed.")
+    typer.echo(summary)
+
+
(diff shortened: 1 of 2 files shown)


eis_toolkit/cli.py
```diff
@@ -3065,7 +3074,149 @@ def gamma_overlay_cli(input_rasters: INPUT_FILES_ARGUMENT, output_raster: OUTPUT
...
+    arrays_to_generate: Annotated[Optional[List[str]], typer.Option()] = None,
+):
+    """
+    Calculate weights of spatial associations.
+
+    Parameter --studentized-contrast-threshold is used with 'categorical', 'ascending' and 'descending' weight types.
+
+    Parameter --arrays-to-generate controls which columns in the weights dataframe are returned as arrays. All column
+    names in the produced weights_df are valid choices. Defaults to ["Class", "W+", "S_W+] for "unique" weights_type
+    and ["Class", "W+", "S_W+", "Generalized W+", "Generalized S_W+"] for the cumulative weight types.
```
nmaarnio, line 3094: This is a pretty horrible parameter for the user, but can't help it when the toolkit defines things as it does now. I think we could at least list the options, both here and in the toolkit function docstring. In QGIS let's figure out something else for this than just a text input

eis_toolkit/cli.py
```diff
@@ -3065,7 +3074,149 @@ def gamma_overlay_cli(input_rasters: INPUT_FILES_ARGUMENT, output_raster: OUTPUT
...
+        evidential_raster=evidential_raster,
+        deposits=deposits,
+        raster_nodata=raster_nodata,
+        weights_type=weights_type,
+        studentized_contrast_threshold=studentized_contrast_threshold,
+        arrays_to_generate=arrays_to_generate,
+    )
+    typer.echo("Progress: 75%")
+
+    df.to_csv(output_dir.joinpath("wofe_results.csv"))
```
nmaarnio, line 3117: Perhaps CSV is a good output format, it can be read into a vector layer without geometries in QGIS

eis_toolkit/cli.py
```diff
@@ -3065,7 +3074,149 @@ def gamma_overlay_cli(input_rasters: INPUT_FILES_ARGUMENT, output_raster: OUTPUT
...
+
+    with rasterio.open(output_probabilities_std_path, "w", **raster_profile) as dst:
+        dst.write(posterior_probabilies_std, 1)
+
+    with rasterio.open(output_confidence_array_path, "w", **raster_profile) as dst:
+        dst.write(confidence_array, 1)
+
+    typer.echo("Progress: 100%")
+
+    typer.echo(f"Results saved in {output_dir}.")
```
nmaarnio, line 3219:
```suggestion
    typer.echo(f"Responses calculations finished, writing output rasters to {output_raster_probabilities},  {output_raster_standard_deviations} and {output_raster_confidence}.")
```
If you change the output dir to output raster paths

nmaarnio (changes requested):
In addition to my comments below, please add CLI function for `agterberg_cheng_CI_test` tool

eis_toolkit/cli.py
```diff
@@ -3065,7 +3087,157 @@ def gamma_overlay_cli(input_rasters: INPUT_FILES_ARGUMENT, output_raster: OUTPUT
...
+        raster_nodata=raster_nodata,
+        weights_type=weights_type,
+        studentized_contrast_threshold=studentized_contrast_threshold,
+        arrays_to_generate=arrays_to_generate,
+    )
+    typer.echo("Progress: 75%")
+
+    df.to_csv(output_dir.joinpath("wofe_results.csv"))
+    for key, array in arrays.items():
+        output_raster_path = output_dir.joinpath(key + ".tif")
```
nmaarnio, line 3135: I would make the names of the output files a bit clearer. Typically the user has to run the weights calculations multiple times as far as I know, so we should avoid duplicate names. One option that came to my mind is `<input_file_name>_weights_<weights_type>_<column>`, for example "wofe_evidence_raster_weights_categorical_W+.tif"

nmaarnio (commented):
I think it looks good now and seems to work. I would only change the output file naming logic, see my comment below

nmaarnio (approved):

PR #476, opened 2024-12-13
added closure checks also when geochemical data that is in ppm or ppb

jtlait:
Added closure checks also when units are ppm or ppb.


nmaarnio (approved):
Looks good to me!

PR #481, opened 2025-01-10
473 add training data tool

okolekar:
Added the code to convert data points to raster. The base raster values will be 0 and the points will have the value as per the attribute selected.


nmaarnio:
Hi @okolekar , what's the status of this PR – is it ready for review?

okolekar:
Yes it is now ready for review.

okolekar:
Hi Niko,
the PR is ready for review.
Thankyou for your time and support.

.vscode/launch.json
nmaarnio, line 1: This file should be removed from the PR

eis_toolkit/training_data_tools/points_to_raster.py
```diff
@@ -0,0 +1,218 @@
...
+    if os.path.exists(path) and overwrite is False:
+        print(f"File already exists: {os.path.basename(path)}.")
+        return
+    else:
+        if array.ndim == 2:
+            array = np.expand_dims(array, axis=0)
+
+    with rasterio.open(path, "w", compress="lzw", **meta) as dst:
+        dst.write(array)
+        dst.close()
```
nmaarnio, line 103: This function should be removed as it does not relate to the implemented feature itself

eis_toolkit/training_data_tools/points_to_raster.py
```diff
@@ -0,0 +1,218 @@
...
+            )
+            raster_array[positives_rows, positives_cols] = values
+
+            unique_values = list(set(values))
+
+            if radius is not None:
+                for target_value in unique_values:
+                    raster_array = _create_buffer_around_labels(raster_array, radius, target_value, buffer)
+
+    return raster_array, raster_meta
```
nmaarnio, line 140: I would refactor this function and leave out the `MemoryFile` stuff.

eis_toolkit/training_data_tools/points_to_raster.py
```diff
@@ -0,0 +1,218 @@
...
+    template_raster: Optional[rasterio.io.DatasetReader] = None,
+    coord_west: Optional[Number] = None,
+    coord_north: Optional[Number] = None,
+    coord_east: Optional[Number] = None,
+    coord_south: Optional[Number] = None,
+    target_epsg: Optional[int] = None,
+    target_pixel_size: Optional[int] = None,
+    raster_width: Optional[int] = None,
+    raster_height: Optional[int] = None,
+    nodata_value: Optional[Number] = None,
```
nmaarnio, line 158: I would harmonize the inputs of this function to match how we have implemented other EIS Toolkit tools. That is, when a base raster is needed, the user needs to give raster profile / metadata and the base raster is constructed from that, see e.g. `distance_computation` implementation. We will provide the user the option to give either a base raster or define pixel size + extent in QGIS if they wish to do so. So I suggest us to remove all parameters from `template_raster` up until `nodata_value` and simply put `raster_profile` parameter there instead

nmaarnio: These ways to define a base raster you have defined here are anyway available as `create_constant_raster_...` tools separately

eis_toolkit/training_data_tools/points_to_raster.py
```diff
@@ -0,0 +1,218 @@
...
+    Provide 3 methods for raster creation:
+    1. Set extent and coordinate system based on a template raster.
+    2. Set extent from origin, based on the western and northern coordinates and the pixel size.
+    3. Set extent from bounds, based on western, northern, eastern and southern points.
+
+    Always provide values for height and width for the last two options, which correspond to
+    the desired number of pixels for rows and columns.
+
+    Args:
+        positives: The geodataframe points set to be converted into raster.
```
nmaarnio, line 172: I think the first parameter could be more generic like `deposits` or just `geodataframe` – unless this tool is meant just for positive point sample, in which case I think the tool's and Python module's names should reflect this more precisely.

nmaarnio (changes requested):
Hi @okolekar . Good work with this one, looks like a well-designed tool! However, the `points_to_raster` tool seems to me a lot  like our `rasterize` tool. The `rasterize` tool allows also users to buffer around points, so I wonder what exactly do we want out of this tool that `rasterize` does not handle? I might be just missing something here. I did not yet have time to look into `random_sampling.py`, I will do that too soon hopefully

.vscode/launch.json
okolekar: Added the file to the git ignore.

okolekar:
> Hi @okolekar . Good work with this one, looks like a well-designed tool! However, the `points_to_raster` tool seems to me a lot like our `rasterize` tool. The `rasterize` tool allows also users to buffer around points, so I wonder what exactly do we want out of this tool that `rasterize` does not handle? I might be just missing something here. I did not yet have time to look into `random_sampling.py`, I will do that too soon hopefully

This tool allows us to select the radius size for the buffer and also allows to select the type of buffer (min max or average) In addition to the above mentioned points it would also be possible (in the future) to select the shape of the buffer to be applied like square or circular if requested.

nmaarnio (changes requested):
Hi Omkar, I took another look at this and I'd just make a small modification to one parameter type. Other than that, one test seems to be failing and the docs files missing but after they are fixed, we can merge this!

eis_toolkit/training_data_tools/points_to_raster.py
```diff
@@ -0,0 +1,160 @@
...
+    return raster_array
+
+
+@beartype
+def points_to_raster(
+    geodataframe: geopandas.GeoDataFrame,
+    raster_profile: Union[profiles.Profile, dict],
+    attribute: Optional[str] = None,
+    radius: Optional[int] = None,
+    buffer: Optional[str] = None,
```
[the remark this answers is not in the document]
okolekar: Hello Niko,
I have made the requested changes!

nmaarnio (approved):
Looks good now, merging!

PR #482, opened 2025-02-04
Wofe CLI fixes

nmaarnio:


nmaarnio (commented):
See my comment below. Also, `agterberg_cheng_CI_test` should also take in the `weights_df` instead of nr of deposits

PR #485, opened 2025-02-25
Improve CLI feedback

mipeso:
I edited the CLI functions so that they now provide user with messages on their progress. I wanted to avoid hardcoding the messages in the CLI functions but at the same time to make sure they are handed out at correct times. Therefore I used the `contextlib.contextmanager`. @nmaarnio Let me know what you think.

### DIFF
--- a/eis_toolkit/cli.py
+++ b/eis_toolkit/cli.py
@@ -16,7 +16,7 @@ import numpy as np
 import pandas as pd
 import rasterio
 import typer
-from beartype.typing import List, Optional, Tuple, Union
+from beartype.typing import List, Optional, Sequence, Tuple, Union
 from typing_extensions import Annotated
 
 from eis_toolkit.utilities.nodata import nan_to_nodata, nodata_to_nan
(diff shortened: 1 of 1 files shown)


nmaarnio (commented):
I think using context managers is a pretty good idea. However, I would make it a bit simpler than in your implementation and get rid of string parameters. I propose the following structure:

I propose using helper classes like these:

 ```
class ProgressLog():

    @contextmanager
    @staticmethod
    def reading_input_files():
        # typer.echo(f"Progress: {progress_at_start}%")
        typer.echo("Opening input files....")
        yield
        typer.echo("✅ Input files read\n")


    @contextmanager
    @staticmethod
    def running_algorithm():
        typer.echo("Running algorithm...")
        yield
        typer.echo("✅ Algorithm run succesfully\n")


    @contextmanager
    @staticmethod
    def saving_output_files():
        # typer.echo(f"Progress: {progress_at_start}%")
        typer.echo("Saving output files...")
        yield
        typer.echo("✅ Output files saved\n")


    @staticmethod
    def finish():
        typer.echo("✅ Algorithm execution finished succesfully\n")

class ResultSender():

    @staticmethod
    def send_dict_as_json(dictionary: dict):
        typer.echo(f"Results: {json.dumps(dictionary)}")
  ```
        
 here is one example cli func with this proposed structure:
 
 ```
 # NORMALITY TEST RASTER
@app.command()
def normality_test_raster_cli(input_raster: INPUT_FILE_OPTION, bands: Optional[List[int]] = None):
    """Compute Shapiro-Wilk test for normality on the input raster data."""
    from eis_toolkit.exploratory_analyses.normality_test import normality_test_array

    with ProgressLog.reading_input_files():
        with rasterio.open(input_raster) as raster:
            data = raster.read()
            if len(bands) == 0:
                bands = None

    with ProgressLog.running_algorithm():
        results_dict = normality_test_array(data=data, bands=bands, nodata_value=raster.nodata)

    ResultSender.send_dict_as_json(results_dict)

    ProgressLog.finish()
```

nmaarnio:
I can take care of adjusting EIS QGIS Plugin to these new messages since I already started tinkering with that

nmaarnio (approved):
Looking good!

PR #502, opened 2025-03-28
473 add training data tool

okolekar:
Radius in the points_to_raster has now the same units as the reference raster_profile.
Added the checks to the attribute parameter in the points_to_raster.py to check the presence of the column in the geodataframe and if all the values in the column are of numerica type.
Added extended description to both the generate_negatives and points_to_raster.py


nmaarnio (approved):
Looks good now, thanks!

PR #504, opened 2025-03-31
473 add training data tool

okolekar:
Added warning: Application of radius to a non square raster is not possible.


nmaarnio:
@okolekar could you create a new pull request where the branch is based on the current master branch? Right now this PR shows the old commits which should not be the case. Even if working on an "old" feature, I suggest creating new branches that are based on up-to-date master branch for the feature/fix.

nmaarnio:
Alternatively, you can try rebasing your current branch, but it might get a bit complicated and creating a new one might just be easier option

okolekar:
Okay may be I saw your comment a minute late. Will create a new pull request.

okolekar:
> @okolekar could you create a new pull request where the branch is based on the current master branch? Right now this PR shows the old commits which should not be the case. Even if working on an "old" feature, I suggest creating new branches that are based on up-to-date master branch for the feature/fix.

Hi Niko,
I will take care regarding this issue in future commits. Thanks for the note!

